PR #81, opened 2020-05-19
Add maps feature with reverse geocoding

MehdiSaffar:
@umutoksuz and I have been working on the maps feature.

When the user comes to the homepage (`/pages/index.js`) they get a welcome screen and a page-wide maps widget (built on [deck.gl](https://deck.gl/)). On the map there is a cursor that shows the user's position. By default the user is located in Hisarustu (needs to be discussed)

The user can select their location either by panning the map and selecting somewhere, or by using the Search place feature in the top left corner. when a place is selected, a request containing the `lat,lng` information is sent to the `api/getLocationInfo` route in our backend.

Our backend proceeds to fetching the location information from the third party API [position stack](https://positionstack.com/). If the location is invalid (for example clicking in the middle of the ocean) then it returns `{valid: false}` otherwise it selects the information about `{country, language, currency}` and returns it.

This is what the webpage looks like after searching for New York City and pressing Enter

image

You can find the unit tests for the location service under `/services/getLocationInfo.spec.js`. Currently all unit tests work:

```
(ox)  practice_app (maps) npm run test    

> template_next@0.1.0 test /home/mehdisaffar/Documents/courses/cmpe352/bounswe2020group2/practice_app
> jest --passWithNoTests

  processLocationInfo
    ✓ returns valid on valid input (2 ms)
    ✓ returns invalid on no country
    ✓ returns invalid on no currencies (1 ms)
    ✓ returns invalid on bad data shape
    ✓ returns the country from the input
    ✓ contains the property currency: {symbol, code, name} (1 ms)
    ✓ uses the first currency in the array
    ✓ always contains the boolean property "valid"

Test Suites: 1 passed, 1 total
Tests:       8 passed, 8 total
Snapshots:   0 total
Time:        0.691 s, estimated 1 s
Ran all test suites.      
```


umutoksuz (commented):
I have worked with @MehdiSaffar on this maps task. Our demo website shows user a map from MAPBOX api and user can select any location on whole earth. After selection we show user which county is selected, what language is spoken in there and currency of that place with use of positionstack api. Functionality of our api works fine and all unit tests are passed.

practice_app/pages/index.js
```diff
@@ -1,9 +1,13 @@
 import { useState } from 'react'
```
furkannane, line 1: Currently map is responding only once. This needs to be fixed.

umutoksuz: It is because missing api key for positionstack. If you create .env file in project directory and paste 

POSITION_STACK_API_KEY=<edited_out>

into that file, it will work fine

furkannane: I did that and it does work fine but i am just curious about how the instructors are going to be aware of this or will you give them this file somehow?

MehdiSaffar: I am not sure, I think API keys should not be pushed to github. But perhaps we share the API keys over slack so that whenever someone wants to test they can put them in an `.env` file and run the app?

practice_app/components/MapView.module.css
```diff
@@ -1,5 +1,5 @@
 .mapContainer {
     position: relative;
-    width: 100vw;
```
MehdiSaffar: why remove width? I think it looks cool when it wide all over the page :D

umutoksuz: Ok, we can set width again. We only apply margin to text part then.

MehdiSaffar: Okay can you revert the width change please? Then we can accept this PR I think

practice_app/pages/index.js
furkannane: I actually have no idea what to do about this. Sharing the key over Slack would make our group be able to run the app but will we share it with the instructors aswell and if we will, then how are we going to do it?

MehdiSaffar: When deploying it I believe we can pass the environment variables (api key) to Docker somehow, so it won't be a problem when demo-ing

PR #83, opened 2020-05-21
Furkan nane covid-19 statistics pull request

furkannane:
In this branch i added statistics functionality. Whenever a user selects a location from the map and clicks the appropriate button, i get the country information from map and i search for istatistics about that country by using a third party api. Then i show the statistics if i am able to find it, otherwise i print "no information found about your country". 

Test are at getStatistics.test.js and it passes all tests. Here is the output:

> template_next@0.1.0 test C:\Users\furka\Documents\GitHub\bounswe2020group2\practice_app
> jest --passWithNoTests

 PASS  pages/api/getStatistics.test.js

Test Suites: 2 passed, 2 total
Tests:       12 passed, 12 total
Snapshots:   0 total
Time:        1.815 s
Ran all test suites.


practice_app/pages/api/getStatistics.js
```diff
@@ -0,0 +1,65 @@
...
+            {
+                // ...
+                res.statusCode = 200
+
+                const countryName = query.country
+
+                let deathToll1
+                let recovered1
+                let answered1
+                let infected1
```
MehdiSaffar: Please bring these variables inside the `then(response ...` part as they don't belong here.

practice_app/pages/api/getStatistics.js
```diff
@@ -0,0 +1,65 @@
...
+
+                axios
+                    .get(
+                        'https://api.apify.com/v2/key-value-stores/tVaYRsPHLjNdNBu7S/records/LATEST?disableRedirect=true',
+                    )
+
```
MehdiSaffar: Please remove whitespace like this because there is no need for it and makes files longer and hard to read.

practice_app/pages/api/getStatistics.js
```diff
@@ -0,0 +1,65 @@
...
+                        for (i = 0; i < response.data.length; i++) {
+                            if (response.data[i].country == countryName) {
+                                answered1 = 'yes'
+                                deathToll1 = response.data[i].deceased
+                                recovered1 = response.data[i].recovered
+                                infected1 = response.data[i].infected
+
+                                break
+                            }
+                        }
```
MehdiSaffar: I suggest you use `Array.find` with an anonymous function to find the country which matches `countryName`. No need for `for` loops and break and stuff.

So you can `find` then take out `deceased, recovered, infected` out of it.

You can read about it here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

practice_app/pages/api/getStatistics.js
```diff
@@ -0,0 +1,65 @@
...
+    // req object contains whatever information we can get from the request that comes from the client (from the browser)
+    // res object is used tow write our reponse
+    const { query, method } = req
+
+    switch (method) {
+        case 'GET':
+            {
+                // ...
+                res.statusCode = 200
+
```
MehdiSaffar: You are setting the status code twice, once here, and once at the end. Remove this

practice_app/pages/api/getStatistics.js
```diff
@@ -0,0 +1,65 @@
...
+                            answered: answered2,
+                            deathToll: deathToll1,
+                            recovery: recovered1,
+                            infection: infected1,
+                        })
+                    })
+                    .catch(error => {
+                        res.statusCode = 500
+
+                        console.log(error)
```
MehdiSaffar: It is not enough to set `res.statusCode`, you also have to tell `next.js` that you have finished handling the request. I think you can use `res.end` or simply `res.json({})`.

Otherwise, if the backend route raises an error, the frontend will keep waiting for the request pretty much forever

practice_app/pages/api/getStatisticsModule.js
```diff
@@ -0,0 +1,16 @@
+exports.checking = function (val) {
+    let cnull = 'no'
+    let ansnull = 'yes'
+
```
MehdiSaffar: Please use better descriptive names, both for the function name and the variable names. I had to look through the code to understand that `cnull` means `countryIsNull`.

practice_app/pages/api/getStatisticsModule.js
```diff
@@ -0,0 +1,16 @@
...
+        return { cnull, ansnull }
+    }
+
+    if (val.answer == null) {
+        ansnull = 'no'
+        return { cnull, ansnull }
+    }
+
+    return { cnull, ansnull }
+}
```
MehdiSaffar: This function is unnecessarily long, Furkan. This can be a one-liner: 
```js
return { cnull: val.countryN == null, ansnull: val.ansnull == null}
```

I would say get rid of this function and make a function that does more than this.

practice_app/pages/index.js
```diff
@@ -66,6 +70,7 @@ export default function Home({ context }) {
             setCountry({ name: country })
             setLanguage({ name: language })
             setCurrency(currency)
+            setCountr(country)
```
MehdiSaffar: Why not use the existing `country` ?

practice_app/pages/index.js
```diff
@@ -33,6 +33,10 @@ export default function Home({ context }) {
         latitude: 41.085212,
     })
 
+    const [numb, setNumb] = useState()
+
+    const [countr, setCountr] = useState()
+
```
MehdiSaffar: Please use descriptive names.

practice_app/pages/index.js
```diff
@@ -75,6 +80,22 @@ export default function Home({ context }) {
...
+        if (data.answered == 'yes') {
+            setNumb(
+                `Country: ${countr} Death Toll:  ${data.deathToll} Recovered People:  ${data.recovery} Infected People: ${data.infection}`,
+            )
+        } else {
+            setNumb('No information available about your country')
+        }
+    }
+
```
MehdiSaffar: 1. Use a better name. This shouldn't be called `getNumber`, should be called something like `getCovidStats`
 2. The way you are doing it is fine, but a cleaner approach would be to `setCovidData({country, deathToll, recovery, infection})` and leave the "presentation formatting" when you are rendering.

practice_app/pages/index.js
```diff
@@ -94,6 +115,11 @@ export default function Home({ context }) {
                 <h1>Welcome to our demo website!</h1>
                 {country && language && currency && <p>{locationGreeting()} </p>}
                 <MapView onMapClick={onMapClick} coordinates={coordinates} />
+
+                <button onClick={getNumber}>
+                    Click me to get Covid death statistics for the country you have chosen on the map
+                </button>
```
MehdiSaffar: This is fine but what do you think about removing the need for a button press, and load the covid stats whenever the user changes location? This can be done by modifying the `onMapClick` function.

practice_app/pages/api/getStatistics.test.js
```diff
@@ -0,0 +1,23 @@
+const { checking } = require('./getStatisticsModule')
+
+describe('getStatistics tests', () => {
```
MehdiSaffar: describe string should contain the exact name of the function you are testing, in my opinion. So it should `describe('checking')`

MehdiSaffar (changes requested):

practice_app/pages/api/getStatistics.js
furkannane: Thats a good suggestion but in terms of complexity i suppose they are the same, if so i would rather not change the code but next time i would use it for sure.

practice_app/pages/index.js
furkannane: i have no idea. I will use that.

practice_app/pages/index.js
furkannane: Honestly i prefer having a front-end part so my code is longer. I barely have enough amount of code so anything that i can add is plus for me.

practice_app/pages/api/getStatistics.js
furkannane: done

practice_app/pages/api/getStatistics.js
furkannane: done

practice_app/pages/api/getStatistics.js
furkannane: done

practice_app/pages/api/getStatistics.test.js
furkannane: done

practice_app/pages/index.js
furkannane: done the 1st one.

practice_app/pages/index.js
furkannane: done

practice_app/pages/api/getStatisticsModule.js
furkannane: I got rid of it and created a better function. Here is unit tests of this new function:
> template_next@0.1.0 test C:\Users\furka\Documents\GitHub\bounswe2020group2\practice_app
> jest --passWithNoTests

 PASS  pages/api/getStatistics.test.js

Test Suites: 2 passed, 2 total
Tests:       12 passed, 12 total
Snapshots:   0 total
Time:        1.953 s
Ran all test suites.

C:\Users\furka\Documents\GitHub\bounswe2020group2\practice_app>

practice_app/pages/api/getStatistics.js
MehdiSaffar: The complexity is indeed the same. Using `map`, `find`, `reduce` etc is often more readable and more idiomatic Javascript. Next time keep it in mind ;)

practice_app/pages/api/getStatisticsModule.js
furkannane: done

practice_app/pages/index.js
MehdiSaffar: If you want more code on the frontend, my suggestion would be to show the numbers using different colors. If you had implemented my second suggestion about `setCovidStat` earlier, you would have been able to wrap the numbers inside a `<span style={{color: 'red'}}>{data.infected}</span>` to show the infected count in red for example, and recovered in green etc..

practice_app/pages/index.js
```diff
@@ -33,9 +33,7 @@ export default function Home({ context }) {
         latitude: 41.085212,
     })
 
-    const [numb, setNumb] = useState()
-
-    const [countr, setCountr] = useState()
+    const [returnString, setReturnString] = useState()
```
MehdiSaffar: I still don't like this name. When someone else on our team comes and wants to add some feature here, they'll have no idea what `returnString` here mean. Please change it to something like `covidInfo` or `covidStats` or something like that.

practice_app/pages/index.js
```diff
@@ -80,19 +77,19 @@ export default function Home({ context }) {
         }
     }
 
-    const getNumber = async () => {
-        if (countr == null) {
-            setNumb('Please select a location by clicking on the map')
+    const getCovidStats = async () => {
+        if (country.name == null) {
```
MehdiSaffar, line 81: This will crash if `country` itself is undefined

MehdiSaffar (changes requested):
In general I approve.

omerfdeniz:
Previously decided features are implemented well. Some variable and functions may be named better, but good job overall

practice_app/pages/index.js
furkannane: But country is already defined globally in this lines: const [country, setCountry] = useState({
        code: 'TR',
        name: 'Turkey',
    })

practice_app/pages/index.js
furkannane: changed the name.

practice_app/pages/index.js
MehdiSaffar: You are right. I had in mind some earlier code where `setCountry(undefined)` could happen.

MehdiSaffar (approved):
I tested it now and it works fine. We can merge it. Good job

PR #86, opened 2020-05-23
Added the create product feature.

asliayk:
Got the values of variables in request body to create a product constant. Specified the resource path. CreateRequest object is initialized and sent it to createProduct function as a parameter. Specified the path of the product and the product set. Created addRequest with these paths and sent it to addProductToProductSet function as a parameter.


karalleyna:
Everything is fine. We can merge it.

karalleyna (commented):
It can create product as intended.

PR #87, opened 2020-05-23
Add create product set function

karalleyna:
After the resource path of the Google Cloud Platform is found,  request consisting of path, the name and id of the product set is sent via createProductSet function of Google Product Search API.


asliayk (approved):
Looks fine. Ready to merge.

PR #88, opened 2020-05-23
Added the add reference image feature.

asliayk:
Got the values of variables in request body to create a reference image constant. Specified the product path. Request object is initialized and sent it to createReferenceImage function as a parameter.


karalleyna (approved):
I think it is complete.

PR #90, opened 2020-05-23
Add update product function

karalleyna:
It is helpful to update labels of product whenever end user wants to change one of the label of the product. The product client of Google Product Search API is called via it's updateProduct function. Since the product has already created, it's request method must be 'PATCH'


asliayk (approved):
I think it's ready to merge.

PR #91, opened 2020-05-23
Reddit branch merge

### LINKED ISSUE #104: Implement Reddit Functionality
Write a basic module for collecting posts from a given subreddit.
Use the selected country name as the selected subreddit.
Implement a button for a really basic UI.
Set up tests for the reddit module.

bcetin:
I added the route /api/getRedditData to our API.

This route collects titles and permalinks of top 10 posts from a given subreddit name.

It accepts a GET Request with a single parameter:
country: Name of the country 

It responds with a single array of information:

posts[]: Each element in posts contains:
- title: Title of the post
- permalink: Permalink of the post.

I added unit tests for the module useing a local example returned value from a request I made manually to the reddit API.

On the index.js page you can click a button and get 10 posts for the selected country.


MehdiSaffar:
- There is no description. Please look at the previous pull requests and use a similar structure
- Did you actually run the tests before submitting this pull request? Just looking at the code I know that it won't pass.
- One test is not enough
- Do not make HTTP requests inside the unit tests as they slow down the entire unit test suite. 
- There is a lot of whitespace all over the place. Please format your code before submitting. Use `npm run util:prettier` for that from the command line.
- There is a lot of code duplication for the posts part, both in the render and inside the route. I showed you on Slack how to use `Array.map`. Please use that.

asliayk:
- The countries that doesn't have reddit info returns nothing. I think it would be better if it informed the user like "no info for that country", as in Furkan's covid API, or maybe as an alert.
- I couldn't run your test as Mehdi couldn't. Maybe you installed something and ran the test after that.

karalleyna (changes requested):
- If the end user chooses somewhere,e.g Black Sea,Atlantic Ocean etc., instead of a country by mistake, the program gives 
`Unhandled Rejection (TypeError): Cannot read property 'name' of undefined`. This can be handled by adding if statement which checks whether countryName is undefined or not after you get the countryName.
- The test does not work properly and it can be better if you use the module which you export in order to test.

asliayk (commented):
- The countries that don't have reddit info returns nothing. I think it would be better if it informed the user like "no info for that country", as in Furkan's covid API, or maybe as an alert.
- I couldn't run your test as Mehdi couldn't. Maybe you installed something and ran the test after that.

MehdiSaffar (changes requested):
(Ignore this comment)

practice_app/pages/api/getRedditData.test.js
```diff
@@ -0,0 +1,17 @@
+const { checkGetRedditData,getPostData  } = require('./getRedditDataModule')
+const testResponse = require('./exampleRedditData.json')
+
+describe('checkGetRedditData', () => {
```
ozdenizdolu, line 4: I think existing tests are good tests but I suggest you also add some tests for a malformed input data. If I understand it correctly, you check `getPostData` function for a well formed json input in `./exampleRedditData.json`. Maybe you can add 2-3 more cases where different parts of the input is malformed or missing.

practice_app/pages/api/getRedditDataModule.js
```diff
@@ -0,0 +1,9 @@
+exports.getPostData = function (response) {
+    const values = response.data.children
+    var outJSON = []
```
ozdenizdolu, line 3: Not a big deal but, I believe using `var` keyword is outdated when declaring variables. You should use `let` and  instead.

ozdenizdolu (approved):
I believe @asliayk  @karalleyna and  @MehdiSaffar has covered almost everything. Here are my comments. I approve the merge and leave considering my comments to you.

furkan-varol:
Some country names gathered from getLocationInfo causes problems when given to reddit api. Like 'United States' causes {"message": "Not Found", "error": 404} for reddit. It must be converted to 'USA' before passing to reddit.

PR #93, opened 2020-05-23
Added the filter products feature.

asliayk:
With the help of listProductsInProductSet function, it is useful to check the categories of each product and list the products in a category on a different page.


karalleyna (approved):
It is ready to merge.

PR #94, opened 2020-05-23
Added the list product sets feature.

asliayk:
With the help of listProductSets feature, it is useful to get and list all set names on a different HTML page.


karalleyna (approved):
I think it's name should be changed and then we can merge it.

asliayk:
Renamed now.

PR #95, opened 2020-05-23
Add get all products function

karalleyna:
It lists all products stored in the storage at new window.


asliayk (approved):
ready to merge.

PR #96, opened 2020-05-23
Add get products in set function

karalleyna:
Products in the product set whose name is given by end user are listed at the new page.


asliayk (approved):
looks fine and ready to merge.

PR #97, opened 2020-05-23
Add delete product function

karalleyna:
It gets the id of the product from request object and deletes it using deleteProduct function from Google Product Search API. Status codes are determined according to possible errors in the documentation of Google Product Search API.


asliayk (approved):
can be merged.

PR #98, opened 2020-05-23
Added delete set feature.

asliayk:
Got the value of product set id in request body, and the product set is deleted with the help of deleteProductSet function.


karalleyna (approved):
Good job! We can merge it

PR #101, opened 2020-05-23
Covid-19 Top 10 statistics pull request

omerfdeniz:
Added Top 10 Statistics Functionality. Here I parse data from a 3rd party API and show the top 10 countries with max number of deaths in a bar chart using chart.js. I tested whether the response that will be shown in the chart consists of 10 country names and 10 death numbers. Here are the test results

jest --passWithNoTests

PASS services/getConvertedPrice.spec.js
PASS services/getLocationInfo.spec.js
PASS pages/api/getStatistics.test.js
PASS pages/api/getCountriesWithMaxDeaths.test.js

Test Suites: 4 passed, 4 total
Tests: 28 passed, 28 total
Snapshots: 0 total
Time: 2.715 s
Ran all test suites.


furkannane (changes requested):
I am getting an error. Map is not displayed properly in this branch.

furkannane (approved):
As far i can see, there is no problem with this branch, it works properly and all tests are passed.

karalleyna (changes requested):
There is a point I'd like to raise before the completion of Covid API. I am still getting `Unhandled Rejection (TypeError): Cannot read property 'name' of undefined` whenever I click somewhere rather than any country. I think the reason for this is trying to get the name of the country variable which is not defined.  I know it is not related to your part. Could you take a look at @furkannane's part again?

umutoksuz (changes requested):
I have ran the app and tested it. For top 10 countries it works fine and shows the chart nicely. But I have also noticed the "Unhandled Rejection (TypeError): Cannot read property 'name' of undefined" error when I change the location from the map and try to get covid statistics of that country. Can you fix it?

asliayk (changes requested):
Since @furkannane 's pull request is closed, I also want to mention some points before merging all branches of Covid API into master.
- For many countries, program resolves without sending a request and says: `API resolved without sending a response for /api/getStatistics?country=France, this may result in stalled requests.` We've also encountered with this problem while testing our code and solved it by initializing promise and resolve.
- For some points on map program returns "no info for that country" or "NA" for a specific key information of that country, but for some other points it returns null and writes it to the page: `France Death Toll: null`. I think it's not a good idea to see different values when an information does not exist on the API.

handesirikci:
it seems to work well and all tests passed successfully

handesirikci (approved):
it seems to work well and all tests passed successfully

furkannane:
Thanks for the issues pointed out by @umutoksuz @karalleyna . Also big thanks to @asliayk for pointing out the issue and the solution. I think i solved the problems: if there is no country in the current location, it doesnt give error. I added promise and resolve. If a value is not found for a statistic, it prints only NA now. Let me know your further toughts.

umutoksuz (approved):

umutoksuz:
I have checked again. Last time I have forgotten to add .env in the project so, I have encountered errors. 
This time it worked as expected and passed all unit tests. I think this covid functionality is ok now.

practice_app/pages/api/getStatistics.js
```diff
@@ -2,44 +2,54 @@ import axios from 'axios'
...
 const { checkApiResponseForCountry } = require('./getStatisticsModule')
 
-export default (req, res) => {
-    // req object contains whatever information we can get from the request that comes from the client (from the browser)
-    // res object is used tow write our reponse
-    const { query, method } = req
+export default (req, res) =>
+    new Promise(async resolve => {
+        // req object contains whatever information we can get from the request that comes from the client (from the browser)
+        // res object is used tow write our reponse
```
MehdiSaffar, line 8: I would suggest that you write `export default async (req, res) => ...` and remove the whole `new Promise(async resolve => { ...` part is it is clearer and does the "Promise wrapping" internally, also handles errors automatically (while with promise you would have to use both `resolve` and `reject` parameters)

practice_app/pages/index.js
```diff
@@ -86,35 +84,94 @@ export default function Home({ context }) {
...
+        const { data } = await axios.get('api/getCountriesWithMaxDeaths')
+        // Print to the front-end
+        const ctx = document.getElementById('bar_deaths')
+        ctx.style.visibility = 'visible'
+        const chart = new Chart(ctx, {
+            // We are creating a bar chart for death statistics of top 10 countries
+            type: 'bar',
+            data: {
+                labels: data.keys,
+                datasets: [
```
MehdiSaffar, line 130: Although this part might work right, mixing pure javascript HTML manipulation with React/Next.js will give hard to solve errors and potentially unpredictable behavior. If you have the time, change this to use `react-chartjs-2` instead. It won't be a big change since you should be able to pass the entire chart configuration directly.

furkan-varol:
passed all tests and everything seems good

MehdiSaffar (approved):

furkan-varol (approved):

PR #105, opened 2020-05-24
Translation API

furkan-varol:
Implemented text translation functionality. The user is requested to type something in any language in the appropriate input box. The language of that input may or may not be explicitly selected by the user. If it is not, Google Translate API can detect the language and returns it in the response anyway. The location(country) of the user is obtained from a 3rd party API by passing the IP address of the user and this information is used to determine our target language of the translation.

I tested for various conditions and all tests are passed.

jest --passWithNoTests
 PASS  pages/api/getStatistics.test.js
 PASS  pages/api/getRedditData.test.js
 PASS  pages/api/getCountriesWithMaxDeaths.test.js
 PASS  services/getConvertedPrice.spec.js
 PASS  pages/api/getTranslation.test.js

Test Suites: 6 passed, 6 total
Tests:       45 passed, 45 total
Snapshots:   0 total
Time:        5.104 s
Ran all test suites.


furkannane:
I selected Bulgaria on the map and tried to translate "How are you". It printed "nasılsın". I think there is a problem. Can you check again if it translates properly? Edit: It was suppose to get the language from ip not from map so it works properly. In this case a clearer text on the button would be nice.

practice_app/pages/api/getTranslation.js
```diff
@@ -0,0 +1,46 @@
...
+    const { query, method } = req
+
+    switch (method) {
+        case 'GET':
+            {
+                console.log(query)
+                let target = query.countryCode == null ? '' : query.countryCode
+                target = target.toLowerCase()
+                const text = query.text == null ? '' : query.text
+                let sourceLang = query.sl == null ? '' : query.sl
```
MehdiSaffar: Unnecessary checks. What happens if sl, text and countryCode is null? Better to check that the input is valid. If input is valid then proceed othewise return with an error.

Also if you are gonna check you should check either null or undefined.

practice_app/pages/api/getTranslation.js
```diff
@@ -0,0 +1,46 @@
...
+                    .get(
+                        `https://translate.googleapis.com/translate_a/single?client=gtx&sl=${sourceLang}&tl=${target}&dt=t&q=${text}`,
+                    )
+                    .then(response => {
+                        const values = checkTranslationResponse(response, sourceLang, target, text)
+                        const responseText = values.translation
+                        sourceLang = values.sourceLang
+                        res.statusCode = 200
+                        res.json({
+                            answered: 'yes',
```
MehdiSaffar: Is it ever possible to have `answered == 'no'` ? If so what is the point? 

Also would have been better to have it as a boolean instead.

practice_app/pages/api/getTranslationModule.js
```diff
@@ -0,0 +1,803 @@
+exports.checkTranslationResponse = function (response, sourceLanguage, targetLanguage, originalText) {
+    const translation = response.data[0][0][0]
+    const sourceText = response.data[0][0][1]
+    const sourceLang = getSourceLangName(response.data[2])
+    const isAlternativeSourceExist = response.data[8] ? response.data[8][0].length > 1 : false
```
MehdiSaffar: What are these numbers for? for example what is response.data[8]?

practice_app/pages/api/getTranslationModule.js
```diff
@@ -0,0 +1,803 @@
...
+
+    const alternativeSourceLangs = []
+    if (isAlternativeSourceExist) {
+        let i
+        for (i = 0; i < response.data[8][0].length; i++) {
```
MehdiSaffar: Declare `let i` inside the initialization of the for loop, so `for(let i = 0; ... )`

practice_app/pages/api/getTranslationModule.js
```diff
@@ -0,0 +1,803 @@
...
+        isAlternativeSourceExist,
+        alternativeSourceLangs,
+        sourceText,
+        originalText,
+        sourceLanguage,
+        targetLanguage,
+    }
+}
+
+var getSourceLangName = (exports.getSourceLangName = function (sl) {
```
MehdiSaffar: Remove the first part `var getSourceLangName = (`

practice_app/pages/api/getTranslationModule.js
```diff
@@ -0,0 +1,803 @@
...
+            break
+        }
+        case 'it': {
+            sourceLangName = 'Italian'
+            break
+        }
+        case 'fi': {
+            sourceLangName = 'Finnish'
+            break
+        }
```
MehdiSaffar: No need for these specific cases since they can be handled by the "default" case. I think you should remove the `switch`  completely.

practice_app/pages/api/getTranslationModule.js
```diff
@@ -0,0 +1,803 @@
...
+    }
+    return sourceLangName
+})
+
+var isoLangs = {
+    ab: {
+        name: 'Abkhaz',
+        nativeName: 'аҧсуа',
+    },
```
MehdiSaffar: This does not belong here and makes the code file too long.

Create a JSON file called languages.json for example and import it like `const languages = require('./languages.json')` at the top of this file.

practice_app/pages/index.js
```diff
@@ -187,6 +192,30 @@ export default function Home({ context }) {
         }
     }
 
+    // This function returns the translated version of the input, to the language spoken in user's location
+    // current functionality : source_lang=english, target_lang=ip_dependent
+    const getTranslation = async () => {
+        // Learn client's ip adress using a third party service
+        const ipinfo = await axios.get('https://json.geoiplookup.io/')
```
MehdiSaffar: Please use camelCase for variable names. `ipInfo`

practice_app/pages/index.js
```diff
@@ -187,6 +192,30 @@ export default function Home({ context }) {
...
+        // Provide client's ip & original text to our api
+        const oText = document.getElementById('originalText').value
```
MehdiSaffar: Please rename this to `originalText` for clarity.

practice_app/pages/index.js
```diff
@@ -187,6 +192,30 @@ export default function Home({ context }) {
...
+        const sourceLang = document.getElementById('sourceLanguage').value
+        const { data } = await axios.get(
+            encodeURI(
+                `api/getTranslation?countryCode=${ipinfo.data.country_code}&text=${encodeURIComponent(
+                    oText,
+                )}&sl=${sourceLang}`,
+            ),
+        )
```
MehdiSaffar: I believe there is no need for `encodeURI` or `encodeURIComponent`. Can you try it without?

practice_app/pages/index.js
```diff
@@ -187,6 +192,30 @@ export default function Home({ context }) {
         }
     }
 
+    // This function returns the translated version of the input, to the language spoken in user's location
+    // current functionality : source_lang=english, target_lang=ip_dependent
+    const getTranslation = async () => {
```
MehdiSaffar, line 199: You should wrap the body of this function in a `try... catch` block because if an error happens it will crash the frontend

MehdiSaffar (changes requested):
Overall it is fine. I would suggest using the `language` information instead of getting the ipinfo everytime. This would allow users to select a different country from the map widget and have the text translated to that country's language.

practice_app/pages/api/getTranslation.js
umutoksuz: Agreed with Mehdi, I think if countryCode, text or source language is null, you should not proceed further. You can directly return an error.

practice_app/pages/api/getTranslationModule.js
umutoksuz: I think if alternative source languages for that texts exists, it is contained in response.data[8].

practice_app/pages/api/getTranslation.js
umutoksuz: I have set source language to spanish, target language to turkish and tried to translate 'process'. And api converted it from english to turkish. Looks like it automatically decides how source language if it is not found in given. Maybe specifying source language can be unnecessary.

umutoksuz (changes requested):
Furkan, I have also tested your code and noticed that no matter which coutry I have selected  on the map my input is translated to Turkish. I think you have a tiny problem in there.

umutoksuz (approved):

practice_app/pages/api/getTranslation.js
furkan-varol: Fixed

practice_app/pages/api/getTranslation.js
furkan-varol: "answered" changed to boolean. It is now false when the input is invalid.

practice_app/pages/api/getTranslationModule.js
furkan-varol: This block is for gathering info from Google Translate response. I think the names of the constants makes it clear.

practice_app/pages/api/getTranslationModule.js
furkan-varol: Fixed.

practice_app/pages/api/getTranslationModule.js
furkan-varol: Fixed.

practice_app/pages/api/getTranslationModule.js
furkan-varol: Done.

practice_app/pages/index.js
furkan-varol: Typo corrected.

practice_app/pages/index.js
furkan-varol: Done.

practice_app/pages/index.js
furkan-varol: This is required to avoid "ERR_UNESCAPED_CHARACTERS" when the input is Chinese, Russian, etc. I think you mean one of them must be unnecessary, however, it just works with both of them.

practice_app/pages/index.js
furkan-varol: Done.

practice_app/pages/api/getTranslation.js
furkan-varol: Fixed.

furkan-varol: I could not reproduce this. When i try, it translates from Spanish and the meaning is same as in English, "surec".

practice_app/pages/api/getTranslationModule.js
furkan-varol: Removed.

vaveylafv:
> Furkan, I have also tested your code and noticed that no matter which coutry I have selected on the map my input is translated to Turkish. I think you have a tiny problem in there.

My initial plan was implementing this feature, so you could be able to select a target language from map for translation. However, api used for location info does not return ISO 639-1 language codes. Thus, I should have found a way to convert language info from location api to ISO 639-1 version, so we can pass it to Google Translate. Then, I thought, in order to present a working translation functionality; the current version is enough(Translating to the locally spoken language of user's whereabouts).

furkannane:
Bu branchi Hüseyyin Seyyidle mi yapıyorsunuz yoksa sen tek mi yapıyorsun?

furkan-varol:
> Furkan, I have also tested your code and noticed that no matter which coutry I have selected on the map my input is translated to Turkish. I think you have a tiny problem in there.

Our initial plan was implementing this feature, so you could be able to select a target language from map for translation. However, api used for location info does not return ISO 639-1 language codes. Thus, we should have found a way to convert language info from location api to ISO 639-1 version, so we can pass it to Google Translate. Then, we thought, in order to present a working translation functionality; the current version is enough(Translating to the locally spoken language of user's whereabouts).

practice_app/pages/index.js
MehdiSaffar: Oh okay I didn't expect non-latin languages to cause this error. Thanks

MehdiSaffar (approved):
It seems there is a merge conflict with practice_app/pages/api/getRedditData.test.js
Once you fix them I think you can merge

furkannane (approved):
Please dont forget to merge master into your branch first and resolve the conflicts carefully. After that you can merge.

furkan-varol:
translation api succesfully merged into master

PR #127, opened 2020-11-07
Initialize Android subrepo

asliayk:
# Type of change
- [x] New feature 

# Description

This pull request initializes the subrepo for android development.

I have added some important dependencies to the project which are for:

- data binding
- navigation
- material components

I have created a splash screen with a gradient background example, an (almost) empty login fragment, and four other empty fragments with the corresponding bottom navigation bar.

I have also created a subfolder for models(data classes) and added an example model for product in it.

I have created two string key files for English and Turkish.

I have written a README.md with instructions on how to run the app for development.

# Screenshots

Not necessary since the fragments are almost empty.

# Test results

No tests.

## Checklist:


ecommerce/android/README.md
```diff
@@ -0,0 +1,8 @@
+# CMPE451 - E-commerce app - Android
+
+# Instructions
+
+1. `git pull`
+2. `cd ecommerce/android/`
+3. Make sure you have installed Android Studio and run the project folder(Getlix) on it.
```
karalleyna: It has a stub Gradle wrapper file, enough to allow for easy import into Android Studio. However, how to build the project from command line can be added into the instructions.
```suggestion
3. Make sure you have installed Android Studio and run the project folder(Getlix) on it. Alternatively, use the `gradlew build` command.
```

karalleyna (commented):
You got off to a flying start in our mobile app. I really like the bottom navigation view and it is enough to have four fragments for the time being. Now, they can be beside the point but we should consider to

1. Provide view models of these four fragments in order to build a mobile app with MVVM pattern.
2. Pick up the layout design of SplashScreenActivity.

ecommerce/android/README.md
asliayk: I updated the README.md now according to your suggestion.

The initial design of the splash screen is temporary, it needs to be updated with a better design. The colors are also temporary since we haven't decided on a color palette as a group yet.

You are right, the view models and necessary adapters should be added in the coming days.

karalleyna (approved):
I think we can now merge this branch because I cannot see any point that is open to change other than the instruction of README.md that was done according to my suggestion and the overall app is sufficient for the sketch of our app.

PR #135, opened 2020-11-12
[FRONTEND] Add Product card component

abdullahyldz:
## Type of change

# Description

Product card components added with name ProductCard.js under src/components folder.
It accepts 3 input parameters :  Title, Price and Rating
Title and Price are necessary, Rating default is 2.5, can be changed and but it can be set to read-only later.

# Screenshots

image

## Checklist:

Unit test is required, I did not write it for now.


ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const ProductCard = ({ Title, Price, Rating = 0 }) => {
+    return (
+        <div>
+            <Card
+                style={{ width: 300 }}
+                cover={
+                    <img
+                        alt="example"
+                        src="https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcS9I14rV_Z52uoBWHbruHsl84HI13mv66EW8A&usqp=CAU"
```
MehdiSaffar: replace the src with "imageUrl" prop

ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
...
+import { Button, Rate } from 'antd'
+import { Card } from 'antd'
+
+const { Meta } = Card
+
+export const ProductCard = ({ Title, Price, Rating = 0 }) => {
+    return (
+        <div>
+            <Card
+                style={{ width: 300 }}
```
MehdiSaffar: I would prefer a "width" prop with default = 300

ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react'
+import { BsPlusCircle } from 'react-icons/bs'
+import { Button, Rate } from 'antd'
+import { Card } from 'antd'
+
+const { Meta } = Card
+
+export const ProductCard = ({ Title, Price, Rating = 0 }) => {
+    return (
+        <div>
```
MehdiSaffar: no need for a root div, Card can be a root component

ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react'
+import { BsPlusCircle } from 'react-icons/bs'
+import { Button, Rate } from 'antd'
+import { Card } from 'antd'
+
+const { Meta } = Card
+
+export const ProductCard = ({ Title, Price, Rating = 0 }) => {
```
MehdiSaffar: use PascalCase for components but camelCase for props

ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
...
+    return (
+        <div>
+            <Card
+                style={{ width: 300 }}
+                cover={
+                    <img
+                        alt="example"
```
MehdiSaffar: optional: replace alt with product title

ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
...
+                        src="https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcS9I14rV_Z52uoBWHbruHsl84HI13mv66EW8A&usqp=CAU"
+                    />
+                }>
+                <Meta description={Price} title={Title} />
+                <Rate allowHalf defaultValue={Rating}></Rate>
+                <span>
```
MehdiSaffar: is the span necessary here? if not can you remove it?

ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react'
+import { BsPlusCircle } from 'react-icons/bs'
+import { Button, Rate } from 'antd'
+import { Card } from 'antd'
+
+const { Meta } = Card
+
+export const ProductCard = ({ Title, Price, Rating = 0 }) => {
+    return (
```
MehdiSaffar: can you wrap the entire root with a 
```js
<Link to={`/products/${productId}`}>...</Link>
```
the URL formats are not set in stone, they will probably change but they do not depend on backend so we are free to do however we want.
for this you will need to add `productId` as a prop

ecommerce/frontend/src/components/ProductCard.js
MehdiSaffar: actually, it would be easier for the parent component to pass an entire product as a parameter instead of extracting every single prop.
right now to pass the information one would need to use `ProductCard` like `<ProductCard title={product.title} price={product.price} rating={product.rating}/>`
the easier way would be to use it like `<ProductCard product={product}/>`
of course there is the possiblity of leaving it like it is right now and use "..." spread operator like `<ProductCard {...product}>` but I wouldn't recommend it!

ecommerce/frontend/src/components/ProductCard.js
```diff
@@ -0,0 +1,27 @@
...
+                cover={
+                    <img
+                        alt="example"
+                        src="https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcS9I14rV_Z52uoBWHbruHsl84HI13mv66EW8A&usqp=CAU"
+                    />
+                }>
+                <Meta description={Price} title={Title} />
+                <Rate allowHalf defaultValue={Rating}></Rate>
+                <span>
+                    <Button icon={<BsPlusCircle />}> Add to cart!</Button>
```
MehdiSaffar: I would say remove the ! mark because imagine with multiple products it would look a bit too much enthusiastic :D

ecommerce/frontend/src/components/ProductCard.js
MehdiSaffar: also since price will be a singular decimal value like "49.99" I would suggest asking for a currency parameter since it is possible to have different currencies.
then you would show {price} {currency} in the render

MehdiSaffar (changes requested):
the product card looks really good in the screenshot, I have suggested a few changes for the component. you did a good job, thanks a lot.

when you finish the requested changes, please notify me to review again. if accepted I will merge it to master.

EDIT1: I just noticed the icon looks a bit odd, could you try to fix it please?

image

ecommerce/frontend/src/components/ProductCard.js
abdullahyldz: Yes, sending Product object will be better and easier. I created dummy product object : You can also use them while testing 
`const products = [
    {
        title: 'Yellow Flowers',
        price: '14.99',
        currency: '₺',
        productId: 1,
        rating: '2.5',
        width: 300,
        imageUrl:
            'https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcS9I14rV_Z52uoBWHbruHsl84HI13mv66EW8A&usqp=CAU',
    },
    {
        title: 'Blue Toy Car',
        price: '23.99',
        currency: '€',
        productId: 2,
        rating: '3.5',
        width: 400,
        imageUrl: 'https://i.ebayimg.com/images/g/2GgAAOSw~PVcpcby/s-l400.jpg',
    },
]`
For now, currency operations are not done in frontend, FE accepts the value and the currency as input and just renders them.

ecommerce/frontend/src/components/ProductCard.js
abdullahyldz: I did not understand this part : 

> entire root with a ... where "path_to_product" . Actually, whole review did not ring a bell. :(

ecommerce/frontend/src/components/ProductCard.js
abdullahyldz: > the product card looks really good in the screenshot, I have suggested a few changes for the component. you did a good job, thanks a lot.
> 
> when you finish the requested changes, please notify me to review again. if accepted I will merge it to master.
> 
> EDIT1: I just noticed the icon looks a bit odd, could you try to fix it please?
> 
> image

Regarding your button review : 
React-icons and antd-button does not fit vertically well. I switched to antd-icons and look good. 
image

MehdiSaffar: perfect!

ecommerce/frontend/src/components/ProductCard.js
MehdiSaffar: thanks a lot!

ecommerce/frontend/src/components/ProductCard.js
MehdiSaffar: sorry some of the react tags did not show in the comment, I have edited it now

ecommerce/frontend/src/components/product_card/ProductCard.js
```diff
@@ -0,0 +1,24 @@
...
+
+    return (
+        <Link to={`/products/${productId}`}>
+            <Card style={{ width: width }} cover={<img alt={title} src={imageUrl} />}>
+                <Meta description={price + ' ' + currency} title={title} />
+                <div className="rate-and-addCart">
+                    <Rate allowHalf defaultValue={rating}></Rate>
+                    <Button type="primary" icon={<PlusCircleOutlined onClick={onAddToCart(productId)} />}>
+                        Add to Cart
+                    </Button>
```
MehdiSaffar: It would be better to have the onClick apply to the whole button instead of only the icon

ecommerce/frontend/src/index.js
```diff
@@ -3,15 +3,37 @@ import ReactDOM from 'react-dom'
...
 
 ReactDOM.render(
-    <React.StrictMode>
-        <App />
-    </React.StrictMode>,
+    <div>
+        {products.map(product => (
+            <ProductCard product={product} onAddToCart={() => {}}></ProductCard>
+        ))}
+    </div>,
```
MehdiSaffar: I think you were testing and forgot to remove this part back

MehdiSaffar (changes requested):

ozdenizdolu:
When I try to run this on my local, I am getting an error as follows:
Error: Invariant failed: You should not use `<Link>` outside a `<Router>`

ecommerce/frontend/src/index.js
abdullahyldz: To checkout to your branch, I needed to commit my changes. Stashing was another option. You disregard this file while merging, or let me check if I can revert this file.

ecommerce/frontend/src/components/product_card/ProductCard.js
abdullahyldz: Ah, it should've been that way.

ozdenizdolu (dismissed):

ecommerce/frontend/src/index.js
MehdiSaffar: It would be better if you revert this file, you can even copy directly from master now :D

MehdiSaffar:
> When I try to run this on my local, I am getting an error as follows:
> Error: Invariant failed: You should not use `<Link>` outside a `<Router>`

This is only an issue because this branch doesn't have the routing stuff from my PR and only because we are using the ProductCard component. It will work fine once a ProductCard is actually used after my PR (which is accepted now btw)

ecommerce/frontend/src/components/product_card/ProductCard.less
```diff
@@ -0,0 +1,15 @@
...
+    justify-content: space-between;
+    column-count: 2;
+    margin-top: 7px;
+}
+
+.left {
+    // float: left;
+    margin-right: auto;
+}
+.rigth {
```
MehdiSaffar: you have a typo here :D

ecommerce/frontend/src/components/product_card/ProductCard.js
```diff
@@ -0,0 +1,29 @@
...
+            <Card style={{ width: width }} cover={<img alt={title} src={imageUrl} />}>
+                <Meta description={price + ' ' + currency} title={title} />
+                <div className="rate-and-addCart">
+                    <div className="left">
+                        <Rate allowHalf defaultValue={rating}></Rate>
+                    </div>
+                    <div className="right">
+                        <Button type="primary" icon={<PlusCircleOutlined />} onClick={b => onAddToCart(productId)}>
+                            Add to Cart
+                        </Button>
```
MehdiSaffar: If you don't need the `b` parameter you can always write `()` instead. This is not important, you can ignore this suggestion.

Beauty of working with javascript (instead of python haha)

MehdiSaffar (approved):

PR #136, opened 2020-11-13
[FRONTEND] Initial layout, global state management, routing, and a few more

MehdiSaffar:
## Type of change

# Description

- Initial layout skeleton with (check out `src/components/layout/`:

  - A header (navbar) containing logo, login and signup button
  - A main content (for now a placeholder)
  - A footer with placeholder content

- Added `src/common.less` containing:

  - color palette as constants

  it should be imported whenever you need it.

- A basic routing mechanism in `src/components/layout/Content.js`

  - /login brings to login page
  - /signup brings to singup page
  - any other route brings to home page

- Replace default page title and favicon with Getflix slogan and logo

- Create initial AppContext to manage global state using `React.Context` and `constate` 3rd party library.

  - Currently contains user/setUser as an example

# Screenshots

### Home page

image-20201113140224531

All other pages just show a string as a place holder

# Test results

No units tests.

## Checklist:

  Code is fairly self-explanatory so no need for clutter with comments.

- URL routes need to discussed
- Currently routes are not protected (shouldn't show Login and Signup when user is logged in for example)
- Needs to be tested with actual content (not sure about how the footer will behave for example, I hope it stays sticking at the bottom, kinda had to hack it)


ecommerce/frontend/src/components/CategoryBar.js
```diff
@@ -0,0 +1,17 @@
+import { Menu } from 'antd'
+import { Link } from 'react-router-dom'
+
+export const CategoryBar = ({ categories = [] }) => {
```
abdullahyldz, line 5: Although it has no problem, in the future, we may need to add sub sub categories, so a dropdown feature may be added to Category bar.

ecommerce/frontend/src/components/CategoryBar.less
```diff
@@ -0,0 +1,5 @@
+
+.category-menu {
```
abdullahyldz, line 2: I think this css needs to be in HomePage.less to center the CategoryBar 
image

Centered categories look better to me.

ecommerce/frontend/src/components/layout/Layout_common.less
```diff
@@ -0,0 +1,34 @@
...
+    font-size: 2em;
+}
+
+.header {
+    padding-left: 0px;
+    padding-right: 0px;
+    position: 'fixed';
+    z-index: 1;
+    width: 100%;
+    background: @Eggplant;
```
abdullahyldz, line 16: Instead of Eggplant, we may choose softer colors for header, and 1-bit softer color for category bar. Just an idea, we can discuss later.

abdullahyldz: Also, if we show login and signup buttons at the header, they need to have the same style as the whole header. 
image

abdullahyldz: I could not make a comment on favicon.ico, I do it here : 
Putting logo on tab section is a good idea, however, since the icon is pretty dark. It doesn't look good. 
Later, we can open issue and redesign the coloring of our icon, for this case. 
I think, it is a good idea for us to have 2 different colored versions of our logo. 
Black background + Champagne font
Champagne font + Black font

ozdenizdolu (approved):

ecommerce/frontend/src/components/CategoryBar.js
ozdenizdolu: I thinks it's a good consideration but it's not immediately needed as you have said.

ecommerce/frontend/src/components/layout/Layout_common.less
ozdenizdolu: Yes I also noticed that the favicon looks too dark as it is right now. It's good that you have changed it to getflix logo rather than react default icon, but we should definitely state this to the group in the next meeting.

ecommerce/frontend/src/components/CategoryBar.less
MehdiSaffar: by centering do you mean vertically or horizontally? because I believe they already are centered horizontally.

ecommerce/frontend/src/components/layout/Layout_common.less
MehdiSaffar: > Also, if we show login and signup buttons at the header, they need to have the same style as the whole header.

do you mean the entire bar should be the same color? so like either all white or all that color?



> I could not make a comment on favicon.ico, I do it here :
> Putting logo on tab section is a good idea, however, since the icon is pretty dark. It doesn't look good.
> Later, we can open issue and redesign the coloring of our icon, for this case.
> I think, it is a good idea for us to have 2 different colored versions of our logo.
> Black background + Champagne font
> Champagne font + Black font



> Yes I also noticed that the favicon looks too dark as it is right now. It's good that you have changed it to getflix logo rather than react default icon, but we should definitely state this to the group in the next meeting.

I agree with you guys, I preferred to leave color selection and style changes for later so that we all agree on it. This pull request was mainly for structural purposes and having something on the screen.

ecommerce/frontend/src/components/CategoryBar.js
MehdiSaffar: We will have to discuss this further since the requirements mention 17 categories. I don't remember seeing subcategories so probably some of them will become subcategories? I am not sure.

We can leave this for another PR.

ecommerce/frontend/src/components/CategoryBar.less
abdullahyldz: I meant the CategoryBar to be horizontally aligned. 
image
Your CSS is in the correct position, but copying it to HomePage.less did what I wanted. I was lucky, I guess.

abdullahyldz (approved):

ecommerce/frontend/src/components/CategoryBar.less
MehdiSaffar: Oh okay, must have happened while splitting the LESS code into different files, I'll fix that. Thanks!

MehdiSaffar:
After taking into consideration feedback from my team, I am squashing&merging to master.

PR #137, opened 2020-11-13
[FRONTEND] Add the signup form component

ozdenizdolu:
## Type of change

# Description

Added `SignUpForm.js` to the folder `./src/components/`. 

This component renders a standard signup form that is compatible with our design. The component has only one prop which is `onSubmit`. The component sends all the data it received to the function given to `onSubmit` prop as a single properly labeled JSON object.

The component checks the validity of the input but the validity checks that are being done is insufficient.

The component has a minimum height constraint because of the input boxes but it is pretty flexible on the width and it is going to fill the container it is placed in.

# Screenshots

- **The signup form**

Example 1Example 2

- **JSON data that the function onSubmit is going to receive upon form submission**

The data form sends to the onSubmit function

# Test results

No testing has done other than by hand. It provides basic functionality in non-extreme cases.

## Checklist:

- **Get Captcha** button is non-functional. 
- It might be helpful to implement a function to initialize the values of the form from a given data.


ecommerce/frontend/src/components/SignUpForm.js
```diff
@@ -0,0 +1,289 @@
...
+                label="Main Address"
+                rules={[
+                    {
+                        required: true,
+                        message: 'Please enter your address!',
+                    },
+                ]}
+            >
+                <Input />
+            </Form.Item>
```
MehdiSaffar: 1. for keys please use camelCase (example mainaddress -> mainAddress, zipcode -> zipCode). It is what we use in JS.
2.

**1.1.1.2.6** Location (not mandatory for customer and admin users)

The location information is not mandatory for customer users. Therefore the `required: true` needs to depend on whether the Guest user is registering as a Customer or Vendor. This might difficult since required needs to react to value change of a form item so I can help you with it when you are available.

ecommerce/frontend/src/components/SignUpForm.js
```diff
@@ -0,0 +1,289 @@
...
+                            ]}
+                        >
+                            <Input />
+                        </Form.Item>
+                    </Col>
+                    <Col span={12}>
+                        <Button>Get captcha</Button>
+                    </Col>
+                </Row>
+            </Form.Item>
```
MehdiSaffar: Good thinking. Not sure how we will do it but maybe using this library would help https://www.npmjs.com/package/react-google-recaptcha

ecommerce/frontend/src/components/SignUpForm.js
```diff
@@ -0,0 +1,289 @@
...
+                    {
+                        validator: (_, value) =>
+                            value ? Promise.resolve() : Promise.reject('Should accept agreement'),
+                    },
+                ]}
+                {...tailFormItemLayout}
+            >
+                <Checkbox>
+                    I have read the <a href="">agreement</a>
+                </Checkbox>
```
MehdiSaffar: Note to self: need to either ask customer for some agreement document or just generate a generic one.

MehdiSaffar (changes requested):
Overall I like it very much!

Like I said, please have the keys in camelCase and try to see if you can integrate ReCAPTCHA. If you need help we can pair code together. If ReCAPTCHA takes too long we can leave it for another PR.

ecommerce/frontend/src/components/SignUpForm.js
ozdenizdolu: 1. I generally automatically use camelCase but I missed those, ill correct it. 
2. Maybe we can set a `onChange` callback on the radio buttons and define a state like `locationEnabled` and make it both disable the input boxes and set `required: locationEnabled`. Will that work?

ecommerce/frontend/src/components/SignUpForm.js
ozdenizdolu: Let me try to implement this, I'll ask help if I have a hard time

ecommerce/frontend/src/components/SignUpForm.js
ozdenizdolu: corrected the non camelCase stuff on 4th commit

MehdiSaffar: for 2:

That sounds like it could work actually. Can you try it? If it still doesn't work then let me know.

abdullahyldz:
Thanks for the effort, Ozdeniz. Looking good!
I used your component by adding  <SignUpForm></SignUpForm> in the SignupPage.js Is it the correct way of using it?
When I do so, your component and footer css are mixed to each other. 
image
But, when I zoom out, they look ok. 

Also, when I press "Get Captcha" button, my browser page shakes :) This happens only 75% zoom. 

I think, both issues are related to global css.

ecommerce/frontend/src/components/SignUpForm.js
```diff
@@ -0,0 +1,311 @@
...
+
+            {/* Zip code input */} 
+            <Form.Item
+                id="zipCodeInput"
+                // hidden={!addressEnabled}
+                name={['address', 'zipCode']}
+                label="Zip Code"
+                rules={[
+                    {
+                        required: addressEnabled,
```
abdullahyldz: ZipCode Rules: Is there a rule for 'only numeric' input? 
Phone number Rules: Is there a rule for 'certain lengths'? 

Maybe we can improve the form answers.

ozdenizdolu: Mehdi and I worked on that today, Mehdi pushed a corrected version. So they will be controlled.

ozdenizdolu: I think for your previous comment(about the page is being messy and footer floating on top of another): it feels like it is related to the layout. I believe we should tackle that problem when putting everything together.

ozdenizdolu:
Okay guys, thanks for the reviews. I think I have finalized it. If you have time you can check for one last time if you want. The form looks like as follows right now:

ecommerce/frontend/src/components/SignUpForm.js
```diff
@@ -131,159 +167,187 @@ export const SignUpForm = ({ onSubmit = () => { } }) => {
...
+                        rules={[
+                            {
+                                required: true,
+                                message: 'Please enter your country code!',
+                            },
+                            {
+                                message: 'Wrong country code format',
+                                validator: (rule, val) =>
+                                    /\+\d{1,3}/.test(val) ? Promise.resolve(true) : Promise.reject("Country code format is incorrect!"),
+                            },
```
MehdiSaffar: nitpick: if you have time, make it so that the validator returns an error only when both the field is not empty and the value given does not fit the format. this is done so that you dont get both "please enter country code" and "wrong country code format" when the field is empty. this suggestion is optional, we might leave it for another PR. if you are going to do it, do it for all the similar fields.

ecommerce/frontend/src/components/SignUpForm.js
```diff
@@ -131,159 +167,187 @@ export const SignUpForm = ({ onSubmit = () => { } }) => {
...
+                    onOk={onModalAgreementAccept}
+                    onCancel={onModalAgreementReject}
+                    bodyStyle={{height: 500}}
+                    footer={[
+                <Button key="back" type="primary" onClick={onModalAgreementReject}>
+                    I reject
+                </Button>,
+                <Button key="submit" type="primary" onClick={onModalAgreementAccept}>
+                    I accept
+                </Button>,
```
MehdiSaffar, line 373: not sure if you need the callbacks both on onOK and on the button itself. It might make the callback be called twice.

ecommerce/frontend/src/index.js
```diff
@@ -3,11 +3,12 @@ import ReactDOM from 'react-dom'
...
 
 ReactDOM.render(
-    <React.StrictMode>
-        <App />
-    </React.StrictMode>,
+    // <React.StrictMode>
+    <SignUpForm />,
+    // </React.StrictMode>,
     document.getElementById('root'),
 )
```
MehdiSaffar: Replace SignUpForm with App :D

MehdiSaffar (approved):
Overall very good job, thank you for your effort.

I gave a few nitpicks, you are free to do them if you have time.

You can squash and merge to master.

abdullahyldz (approved):
Looking good.

ecommerce/frontend/src/components/SignUpForm.js
ozdenizdolu: I have checked for double callback case. It only calls them once. So I am not going to change anything here.

ecommerce/frontend/src/index.js
ozdenizdolu: Will do.

ozdenizdolu:
I am merging the pull request. Final form of the forms looks like as follows:

PR #140, opened 2020-11-14
[BACKEND] DB models created, Authentication and Permission mechanism added

umutoksuz:
## Type of change

- [x] Structural features

# Description

Database models are added to create our initial structure and relationships between our models are determined. (Foreign Keys. etc.)
Authentication added. Json Web Tokens will be used to authenticate users. And we use role based permissions to distinguish between types of our users.
Also classes are added for hashing and salting passwords.
/login
/register
/api 
endpoints are added for test purposes they are not totally completed yet.

## Checklist:


omerfdeniz:
We have some __str__ methods in the models, we can remove all of them at the beginning, or add to all models for the sake of clarity. Other than that, great job!

furkannane:
I think Umut can merge so that we can start to use the models in our local.

PR #143, opened 2020-11-14
[FRONTEND] Add Search Input component

abdullahyldz:
## Type of change

# Description

Search bar component witch Category selection and input component.

# Screenshots

image

## Checklist:

Real bottleneck will start when we want to integrate this component with the <Products> component.

### DIFF
--- a/ecommerce/frontend/src/components/SearchInput.js
+++ b/ecommerce/frontend/src/components/SearchInput.js
@@ -1,0 +1,48 @@
+import React from 'react'
+import './SearchInput.less'
+import { Select } from 'antd'
+import Search from 'antd/lib/input/Search'
+
+// TODO - Fetch from a common file or backend
+const categories = [
+    { key: 'electronics', title: 'Electronics' },
+    { key: 'home_and_garden', title: 'Home & Garden' },
+    { key: 'personal_care', title: 'Personal Care' },
+    { key: 'furniture', title: 'Furniture' },
+    { key: 'pet_supplies', title: 'Pet Supplies' },
+    { key: 'mens_fashion', title: "Men's Fashion" },
+    { key: 'womens_fashion', title: "Women's Fashion" },
+]
+
+const { Option } = Select
+
+function handleCategoryChange(value) {
+    console.log(`selected ${value}`)
+}
+
+const categoryDropdown = (
+    <Select defaultValue="Electronics" style={{ width: 120 }} onChange={handleCategoryChange}>
+        {categories.map(category => (
+            <Option value={category.title}>
+                {category.title} key = {category.key}
+            </Option>
+        ))}
+    </Select>
+)
+
+const onSearchInput = value => {
+    if (value === '') {
+        return
+    }
+    // Trigger products page
+    console.log(value)
+}
+
+export const SearchInput = () => {
+    return (
+        <div className="search-input">
+            {categoryDropdown}
+            <Search placeholder="Ürün, kategori veya marka ara" onSearch={onSearchInput} enterButton />
+        </div>
+    )
+}
--- a/ecommerce/frontend/src/components/SearchInput.js
+++ b/ecommerce/frontend/src/components/SearchInput.js
@@ -1,7 +1,9 @@
-import React from 'react'
+import React, { useState } from 'react'
 import './SearchInput.less'
-import { Select } from 'antd'
-import Search from 'antd/lib/input/Search'
+import { Select, Input } from 'antd'
+
+const { Option } = Select
+const { Search } = Input
 
 // TODO - Fetch from a common file or backend
 const categories = [
@@ -14,35 +16,38 @@ const categories = [
     { key: 'womens_fashion', title: "Women's Fashion" },
 ]
 
-const { Option } = Select
+export const SearchInput = (width = 120, onSearch) => {
+    const [category, setCategory] = useState('Electronics')
 
-function handleCategoryChange(value) {
-    console.log(`selected ${value}`)
-}
-
-const categoryDropdown = (
-    <Select defaultValue="Electronics" style={{ width: 120 }} onChange={handleCategoryChange}>
-        {categories.map(category => (
-            <Option value={category.title}>
-                {category.title} key = {category.key}
-            </Option>
-        ))}
-    </Select>
-)
-
-const onSearchInput = value => {
-    if (value === '') {
-        return
+    function handleCategoryChange(value) {
+        setCategory(value)
     }
-    // Trigger products page
-    console.log(value)
-}
 
-export const SearchInput = () => {
+    const categoryDropdown = (
+        <Select defaultValue={category} style={{ width: { width } }} onChange={handleCategoryChange}>
+            {categories.map(category => (
+                <Option value={category.title} key={category.key}>
+                    {category.title}
+                </Option>
+            ))}
+        </Select>
+    )
+
+    const onSearchInput = value => {}
+
     return (
         <div className="search-input">
             {categoryDropdown}
-            <Search placeholder="Ürün, kategori veya marka ara" onSearch={onSearchInput} enterButton />
+            <Search
+                placeholder="Search for products, categories and brands"
+                onSearch={value => {
+                    if (value.trim() === '') {
+                        return
+                    }
+                    onSearch(value)
+                }}
+                enterButton
+            />
         </div>
     )
 }
--- a/ecommerce/frontend/src/components/SearchInput.js
+++ b/ecommerce/frontend/src/components/SearchInput.js
@@ -16,7 +16,7 @@ const categories = [
     { key: 'womens_fashion', title: "Women's Fashion" },
 ]
 
-export const SearchInput = (width = 120, onSearch) => {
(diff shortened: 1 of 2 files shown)


ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
...
+function handleCategoryChange(value) {
+    console.log(`selected ${value}`)
+}
+
+const categoryDropdown = (
+    <Select defaultValue="Electronics" style={{ width: 120 }} onChange={handleCategoryChange}>
+        {categories.map(category => (
+            <Option value={category.title}>
+                {category.title} key = {category.key}
+            </Option>
```
MehdiSaffar: the key should be an attribute to Option like:

```js
<Option key={category.key} value={category.title}>
...
</Options>
```

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
...
+        ))}
+    </Select>
+)
```
MehdiSaffar: I would have categoryDropdown variable inside the SearchInput component, before the return.

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
...
+
+const onSearchInput = value => {
+    if (value === '') {
+        return
+    }
+    // Trigger products page
+    console.log(value)
+}
```
MehdiSaffar: I would also have this inside the SearchInput. Also make sure to trim value from whitespace because if value is 2 whitespaces the first if will be ignored.

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
...
+
+export const SearchInput = () => {
```
MehdiSaffar: add `onSearch` callback prop to inform the parent component.

ecommerce/frontend/src/components/SearchInput.js
MehdiSaffar: Also have the `handleCategoryChange` a function inside the SearchInput component that updates a 

```js
const [category, setCategory] = useState(someDefaultCategory)
```

and have the Select `defaultValue={category}` since defaultValue is only used on the first render of `Select`

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import './SearchInput.less'
+import { Select } from 'antd'
+import Search from 'antd/lib/input/Search'
```
MehdiSaffar: This is the wrong import (not designed for public use). I believe you can import it alongside the `{ Select }` import.

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
...
+    { key: 'electronics', title: 'Electronics' },
+    { key: 'home_and_garden', title: 'Home & Garden' },
+    { key: 'personal_care', title: 'Personal Care' },
+    { key: 'furniture', title: 'Furniture' },
+    { key: 'pet_supplies', title: 'Pet Supplies' },
+    { key: 'mens_fashion', title: "Men's Fashion" },
+    { key: 'womens_fashion', title: "Women's Fashion" },
+]
+
+const { Option } = Select
```
MehdiSaffar: Move this further app right below the imports please.

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
...
+
+function handleCategoryChange(value) {
+    console.log(`selected ${value}`)
+}
+
+const categoryDropdown = (
+    <Select defaultValue="Electronics" style={{ width: 120 }} onChange={handleCategoryChange}>
```
MehdiSaffar: Have the width be a prop with a default value of 120.

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+    // Trigger products page
+    console.log(value)
+}
+
+export const SearchInput = () => {
+    return (
+        <div className="search-input">
+            {categoryDropdown}
+            <Search placeholder="Ürün, kategori veya marka ara" onSearch={onSearchInput} enterButton />
```
MehdiSaffar: Since right now everything in the website is in English, please change the placeholder to English as well.

MehdiSaffar (changes requested):

ecommerce/frontend/src/components/SearchInput.js
abdullahyldz: when Search component calls `onSearch` function: I call local `onSearchInput`.
is it ok to first call  `onSearchInput` then `onSearch` prop

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -0,0 +1,53 @@
...
+        <Select defaultValue={category} style={{ width: { width } }} onChange={handleCategoryChange}>
+            {categories.map(category => (
+                <Option value={category.title} key={category.key}>
+                    {category.title}
+                </Option>
+            ))}
+        </Select>
+    )
+
+    const onSearchInput = value => {}
```
ozdenizdolu: I am not sure about the purpose of  `onSearchInput` ? Asking only because I am curious. I don't have anything more to say.

ozdenizdolu (approved):

ecommerce/frontend/src/components/SearchInput.js
abdullahyldz: `onSearch` will work for filtering the products according to the input value.

We may have a container, say "PRODUCTS", which will make a call to Backend and get all the products. Then, we display them in the PRODUCTS container. 

When a customer enters input and clicks the search icon ->  `onSearch` will raise an event to PRODUCTS to apply the filtering, according to the input.

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -14,35 +16,38 @@ const categories = [
     { key: 'womens_fashion', title: "Women's Fashion" },
 ]
 
-const { Option } = Select
+export const SearchInput = (width = 120, onSearch) => {
+    const [category, setCategory] = useState('Electronics')
```
MehdiSaffar: This won't work because there is only one prop object/argument being passed to SearchInput. Correct way is:

```js
export const SearchInput = ({width = 120, onSearch}) => {
...
}
```

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -14,35 +16,38 @@ const categories = [
...
-    <Select defaultValue="Electronics" style={{ width: 120 }} onChange={handleCategoryChange}>
-        {categories.map(category => (
-            <Option value={category.title}>
-                {category.title} key = {category.key}
-            </Option>
-        ))}
-    </Select>
-)
+    const categoryDropdown = (
+        <Select defaultValue={category} style={{ width: { width } }} onChange={handleCategoryChange}>
```
MehdiSaffar: Remove the brackets around width:

```js
style={{ width: width }}
// or even shorter
style={{ width}}
```

ecommerce/frontend/src/components/SearchInput.js
MehdiSaffar: @abdullahyldz I think I understand what you are trying to say, but the `onSearchInput` is not referred to in any place in the code? and aren't you already using `onSearch` prop to tell the parent component that a search was performed? I think onSearchInput is not used here.

ecommerce/frontend/src/components/SearchInput.js
```diff
@@ -14,35 +16,38 @@ const categories = [
...
             {categoryDropdown}
-            <Search placeholder="Ürün, kategori veya marka ara" onSearch={onSearchInput} enterButton />
+            <Search
+                placeholder="Search for products, categories and brands"
+                onSearch={value => {
+                    if (value.trim() === '') {
+                        return
+                    }
+                    onSearch(value)
+                }}
```
MehdiSaffar: Might as well write `onSearch(value.trim())` so that any whitespace is removed too.

MehdiSaffar (changes requested):
Good work @abdullahyldz, you can merge once you integrate the changes requested! :+1:

ecommerce/frontend/src/components/SearchInput.js
abdullahyldz: Guys, I am sorry. I mixed `onSearchInput `with `SearchInput `and wrote about the component :) 
I moved the calls inside `onSearchInput ` to inside of Search component and I forgot to delete `onSearchInput `
 ```
         <Search
                placeholder="Search for products, categories and brands"
                onSearch={value => {
                    if (value.trim() === '') {
                        return
                    }
                    onSearch(value)
                }}
                enterButton
         />
```

abdullahyldz: I updated my comment above to a more correct version

PR #146, opened 2020-11-14
[FRONTEND] Login Page

furkan-varol:
## Type of change

# Description

Login page that takes email and password as input. 

# Screenshots

login

## Checklist:

1. Google login method is yet to be implemented. 
2. From a security perspective, handling the user password may be altered.


ecommerce/frontend/src/components/pages/LoginPage.js
```diff
@@ -1,3 +1,107 @@
-export const LoginPage = () => {
-    return 'login'
-}
+import './LoginPage.less'
+import { Form, Input, Button, Row, Col } from 'antd'
+import { Link } from 'react-router-dom'
+import Column from 'antd/lib/table/Column'
+
+export const LoginPage = ({ onSubmit = () => { } }) => {
```
ozdenizdolu: not necessarily a problem but this can be named as LoginForm rather than LoginPage,.

ecommerce/frontend/src/components/pages/LoginPage.js
```diff
@@ -1,3 +1,107 @@
...
+    const onFinish = values => {
+        console.log('Received values of form: ', values)
+        onSubmit(values)
+    }
+
+    const onFinishFailed = errors => {
+        console.log('Finish failed: ', errors)
+    }
+
+    const onLoginAsVendor = () => {
```
ozdenizdolu: Is the purpose of this function is to differentiate between user & vendor? Maybe you can define a state like `const [isVendor, setIsVendor] = setState(false)` set it true on click to the vendor submit button. Then return this value on onSubmit.

ozdenizdolu (approved):

ecommerce/frontend/src/components/pages/LoginPage.js
abdullahyldz: Agreeing with Ozdeniz. Building a separate component LoginForm, then integrating it inside the Login Page is a better practice. 
```
export const LoginPage = () => {
    return <LoginForm />
}
```

ecommerce/frontend/src/components/pages/LoginPage.less
```diff
@@ -0,0 +1,34 @@
+.loginForm {
```
abdullahyldz: Class names are written with **hyphen-separated-strings** style. It makes it easy to read. 
A better example would be -> login-form / login-buttons

ecommerce/frontend/src/components/pages/LoginPage.js
```diff
@@ -1,3 +1,107 @@
...
+                    <Input.Password />
+                </Form.Item>
+
+                <div className="loginButtons">
+                    <Form.Item {...tailLayout}>
+                        <Button className="loginButton" type="primary" htmlType="submit">
+                            Login as User
+                        </Button>
+                        <Button className="loginButton" type="primary" htmlType="button" onClick={onLoginAsVendor}>
+                            Login as Vendor
```
abdullahyldz: What is the difference between Login button html types. Doesn't both send HTTP requests to the server?

I looked at the MDN page to understand more. Submit type sends request and button has no default action, so a function is required. 

One radio element from antd to select login user type and one button to submit form values might be easier. (Ozdeniz had used Radio component, you can copy-paste from his commit)

ecommerce/frontend/src/components/pages/LoginPage.less
```diff
@@ -0,0 +1,34 @@
...
+    display: block;
+    margin-top: 80px;
+    margin-left: auto;
+    margin-right: auto;
+    width: 40%;
```
abdullahyldz: I think, letting the component enlarge as much as possible is a better approach. To restrict the growth, we can pass the width prop from the parent to fix its area.

abdullahyldz (changes requested):

ecommerce/frontend/src/components/pages/LoginPage.js
MehdiSaffar: Agree too.

ecommerce/frontend/src/components/pages/LoginPage.js
MehdiSaffar: I am not sure your solution would work @ozdenizdolu though, because doesn't that require first setting the state then clicking on a Login button? so like 2 steps instead of 1?

ecommerce/frontend/src/components/pages/LoginPage.less
MehdiSaffar: agree, let's stick to conventions.

ecommerce/frontend/src/components/pages/LoginPage.js
MehdiSaffar: @abdullahyldz I believe the "default action" associated with htmlType=submit is blocked by antd so it only serves a semantic purpose like this is the main submit function falan. But I think both here need to be regular buttons with their own callback to differentiate between vendor and customer.

@vaveylafv Also please change User to Customer.

ecommerce/frontend/src/components/pages/LoginPage.less
MehdiSaffar: Yes please remove the width: property to allow the component to take as much space available of the parent component.

MehdiSaffar (changes requested):
Can find my requested changes in previous comments.

MehdiSaffar (approved):
All changes integrated. Good job and thank you for your effort :tada:

furkan-varol:
Change requests are discussed with the team and I implemented them, squashing&merging to master.

PR #158, opened 2020-11-17
[FRONTEND] Trending Products Grid in HomePage

furkan-varol:
## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# Description

Trending products grid in the home page is implemented. Takes 5 product images and display them in a grid designed by frontend team. Hovering on products cause zooming into the product image.

# Screenshots
image

# Test results

No testing is done other than trial by hand.

## Checklist:

What to show above the product images should be designed, because I think it looks raw as it is.


ecommerce/frontend/src/components/TrendingGrid.less
```diff
@@ -0,0 +1,62 @@
...
+    display: grid;
+    grid-template-columns: repeat(16, 1fr);
+    grid-template-rows: repeat(8, 3vw);
+    grid-gap: 12px;
+    width: 80%;
+    margin-top: 40px;
+    margin-left: 10%;
+    margin-right: 10%;
+    margin-bottom: 32px;
+}
```
MehdiSaffar: Please remove the margins and let the parent figure out the proper positioning. Same thing with width..

ecommerce/frontend/src/components/TrendingGrid.less
```diff
@@ -0,0 +1,62 @@
+.gallery {
+    display: grid;
+    grid-template-columns: repeat(16, 1fr);
+    grid-template-rows: repeat(8, 3vw);
```
MehdiSaffar: Why are you using "vw" unit? Those are viewport units. TrendingGrid shouldn't depend on the whole viewport but only its parent's width.
image

ecommerce/frontend/src/components/TrendingGrid.less
```diff
@@ -0,0 +1,62 @@
...
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    display: block;
+}
+
+.gallery__img:hover {
+    -webkit-transform: scale(1.5,1.5);
+    -webkit-transition: all 0.3s ease-in-out;
+}
```
MehdiSaffar: Do not use -webkit specific CSS rules. Remove the "-webkit-" part.

ecommerce/frontend/src/components/TrendingGrid.less
```diff
@@ -0,0 +1,62 @@
...
+    grid-row-start: 5;
+    grid-row-end: 9;
+}
+
+.gallery__item--5 {
+    grid-column-start: 13;
+    grid-column-end: 17;
+    grid-row-start: 5;
+    grid-row-end: 9;
+}
```
MehdiSaffar: I don't understand why there are so many columns and rows? Wouldn't a 3-columned (2fr, 1fr, 1fr) and 2 rowed (1fr, 1fr) grid suffice? This needs to be rethinked.

ecommerce/frontend/src/components/layout/MainLayout.js
```diff
@@ -9,7 +9,7 @@ export const MainLayout = () => {
         <Layout>
             <Header />
             <Content />
-            <Footer />
+            
         </Layout>
```
MehdiSaffar: Please put the footer back when opening a PR.

ecommerce/frontend/src/components/pages/HomePage.js
```diff
@@ -38,5 +39,12 @@ const HomePage_Splash = () => {
...
 const HomePage_MainContent = () => {
-    return <div style={{ height: 100 }}></div>
+    const images = [
+        "https://stylecaster.com/wp-content/uploads/2018/10/bag-storage.jpg",
+        "https://cdn.pixabay.com/photo/2019/07/13/13/42/watch-4334815_960_720.jpg",
+        "https://upload.wikimedia.org/wikipedia/commons/6/65/Product_Photography.jpg",
+        "https://images.unsplash.com/photo-1526947425960-945c6e72858f?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&w=1000&q=80",
+        "https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcQwCP24CO6WEGAUhkfVg4ozqfeve3pvkhEEZg&usqp=CAU"
+    ]
+    return <TrendingGrid imageUrls={images} />
```
MehdiSaffar: It is not enough to just pass imageUrls. The trending products are regular products from backend. Each is an object with an id, pictureUrl etc. accept products instead of imageUrls.

ecommerce/frontend/src/components/TrendingGrid.js
```diff
@@ -0,0 +1,23 @@
...
+            </figure>
+            <figure className="gallery__item gallery__item--3">
+                <img src={imageUrls[2]} className="gallery__img" alt="Image 3" />
+            </figure>
+            <figure className="gallery__item gallery__item--4">
+                <img src={imageUrls[3]} className="gallery__img" alt="Image 4" />
+            </figure>
+            <figure className="gallery__item gallery__item--5">
+                <img src={imageUrls[4]} className="gallery__img" alt="Image 5" />
+            </figure>
```
MehdiSaffar: I would say make this a `{product.map(product => ...)}` instead of iterating over the images manually like this.
I am not sure what is `<figure/>`'s purpose here? Why not a simple `img` tag?
Also you need to handle clicking on the images to bring the user to the product's page. Please use `<Link/>` component from 'react-router-dom`. You can find an example in the ProductCart component.

ecommerce/frontend/src/components/TrendingGrid.js
```diff
@@ -0,0 +1,23 @@
+import './TrendingGrid.less'
+
+export const TrendingGrid = ({ imageUrls = [] }) => {
+    return (
```
MehdiSaffar: replace imageUrls with products

ecommerce/frontend/src/components/TrendingGrid.js
MehdiSaffar: Another reason why turning it into a map is needed is that you are setting imageUrls to be an empty array if no imageUrl is passed but then you assume that imageUrls[0 and 1 and 2 etc until 4] exist..

MehdiSaffar (changes requested):
Quite a few things need to be improved/changed.

ecommerce/frontend/src/components/TrendingGrid.js
ozdenizdolu: Agreed it might be better to assume they are coming from backend. Might be able to implement easily in the future

ozdenizdolu (approved):
I approved but I agree with some of the @MehdiSaffar 's comments

ecommerce/frontend/src/components/TrendingGrid.less
furkan-varol: done

ecommerce/frontend/src/components/TrendingGrid.less
furkan-varol: replaced all with fr

ecommerce/frontend/src/components/TrendingGrid.less
furkan-varol: webkit removed

ecommerce/frontend/src/components/TrendingGrid.less
furkan-varol: done

ecommerce/frontend/src/components/layout/MainLayout.js
furkan-varol: done

ecommerce/frontend/src/components/pages/HomePage.js
furkan-varol: done

ecommerce/frontend/src/components/TrendingGrid.js
furkan-varol: done

ecommerce/frontend/src/components/TrendingGrid.js
furkan-varol: done

ecommerce/frontend/src/components/TrendingGrid.js
```diff
@@ -1,23 +1,24 @@
...
+import { Link } from 'react-router-dom'
+
+export const TrendingGrid = trendingProducts => {
+    var gallerytItemNumber = 0
+    function gallerytItemNo() {
+        gallerytItemNumber += 1
+        return gallerytItemNumber
+    }
+    const products = trendingProducts.trendingProducts
+    console.log(products)
```
MehdiSaffar: please call the parameter `props` instead of `trendingProducts` as that is the convention.

Or you can directly extract the trending products as :
```js
export const TrendingGrid = ({ trendingProducts}) => {
...
}
```

Also please remove the console.log

ecommerce/frontend/src/components/TrendingGrid.js
```diff
@@ -1,23 +1,24 @@
 import './TrendingGrid.less'
+import { Link } from 'react-router-dom'
+
+export const TrendingGrid = trendingProducts => {
+    var gallerytItemNumber = 0
+    function gallerytItemNo() {
+        gallerytItemNumber += 1
+        return gallerytItemNumber
+    }
```
MehdiSaffar: No, please do not use any "hack" like this. The proper way of doing this is by doing 

```js
products.map((product, index) => {
...
})
```

and using `index` for the classNames

ecommerce/frontend/src/components/TrendingGrid.js
```diff
@@ -1,23 +1,24 @@
...
-            <figure className="gallery__item gallery__item--5">
-                <img src={imageUrls[4]} className="gallery__img" alt="Image 5" />
-            </figure>
+            {products.map(product => (
+                <figure className={'gallery__item gallery__item--'.concat(gallerytItemNo())}>
+                    <Link to={'/product/' + product.productId}>
+                        <img src={product.imageUrl} className="gallery__img" alt="" />
+                    </Link>
+                </figure>
+            ))}
```
MehdiSaffar: Great, the only thing missing is using the `product.productId` as a key in `<figure key={product.productId} />` otherwise it will cause problems in the future. I believe React also gives warnings in the console when such a key is missing.

Also check out the previous comment for `galleryItemNo()`

What is figure for? I asked last time but I didn't get a response from you. If it is not necessary it would be better to get rid of it.

ecommerce/frontend/src/components/pages/HomePage.js
```diff
@@ -39,12 +39,60 @@ const HomePage_Splash = () => {
...
+                'https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcQwCP24CO6WEGAUhkfVg4ozqfeve3pvkhEEZg&usqp=CAU',
+            width: 300,
+            productId: 5678,
+        },
     ]
-    return <TrendingGrid imageUrls={images} />
+
+    console.log(products)
+    return (
+        <div className="trending-grid-wrapper">
```
MehdiSaffar: Please remove the console.log

MehdiSaffar (changes requested):
Better but some things still need changes. I like the zoom effect but I think it is a bit too intense. Can you make it zoom a bit less

PR #170, opened 2020-11-19
[FRONTEND] Shopping cart page master component added.

abdullahyldz:
## Type of change

# Description

Shopping cart page master component is created. It will consist of ShoppingCartItems component when @ozdenizdolu creates it. 
This component send current items and their amounts to left component as props. Also, sends onChangeCart function as props to listen to changes in the left part. Then, updates the total price.

# Screenshots

image

# Test results

Please run all the tests and dump the result/summary here in a code block.

## Checklist:

Values of shopping cart, can be obtained from AppContext.js
Layout problems exist.


ecommerce/frontend/src/components/ShoppingCart/ShoppingCartMaster.js
```diff
@@ -0,0 +1,67 @@
+import React, { Component, useState } from 'react'
+import './ShoppingCartMaster.less'
+import { Button } from 'antd'
+import { Statistic, Row, Col } from 'antd'
+
+export const ShoppingCart = ({ currency = 'TL', price = 5, cart }) => {
```
ozdenizdolu: In my opinion, It would be better to get products in the cart as a prop here. Then we can make a request to the backend in the `ShoppingCartPage.js` and send it to this component as a prop. It's just a design choice though.

ozdenizdolu: Now that I think, If we are going to show some type loading icon, it might be better to do the request here and pull the products from inside this component.

ozdenizdolu (dismissed):
I have some comments on aesthetics below, feel free to ignore them:

- I think there is no need for its name to end with Master. 
- I think the black borders look bold, they are kinda killing the vibe. I don't have a better suggestion though :)

I think we cannot merge it as it is because there should be some mockup (but not empty) products that have been sent to ShoppingCartItems to show them. (After I implemented it of course)

ozdenizdolu (commented):
-ignore, this was a duplicate comment-

abdullahyldz:
> I have some comments on aesthetics below, feel free to ignore them:
> 
> * I think there is no need for its name to end with Master.
> * I think the black borders look bold, they are kinda killing the vibe. I don't have a better suggestion though :)
> 
> I think we cannot merge it as it is because there should be some mockup (but not empty) products that have been sent to ShoppingCartItems to show them. (After I implemented it of course)

All agreed.

ecommerce/frontend/src/components/ShoppingCart/ShoppingCartMaster.js
```diff
@@ -0,0 +1,68 @@
...
+    const [totalPrice, setTotalPrice] = useState(price)
+    const [itemCount, setItemCount] = useState(0)
+
+    // When there is any change in the items, call this method with new total price and new item count
+    const onChangeCart = (newPrice, newCount) => {
+        setTotalPrice(newPrice)
+        setItemCount(newCount)
+    }
+
+    // husky test change
```
furkan-varol: irrelevant comment

furkan-varol (approved):

ecommerce/frontend/src/components/ShoppingCart/ShoppingCart.js
```diff
@@ -0,0 +1,88 @@
...
+            } finally {
+                setIsLoading(false)
+            }
+            fetch(`localhost:4000/user/${'userId'}/shoppingCart`).then(res => onChangeCart(res))
+        }
+        // fetch()
+
+        onChangeCart(cart)
+        setIsLoading(false)
+    })
```
MehdiSaffar: It is dangerous to forget the empty dependencies array as a second argument to useEffect as this will make it run everytime the component renders
```js
useEffect(() => {....}, [])
```

MehdiSaffar: I think you shouldn't be calling onChangeCart here
also you are already doing setIsLoading(false) inside the fetch function itself. Please remove this one.

ecommerce/frontend/src/components/ShoppingCart/ShoppingCart.js
```diff
@@ -0,0 +1,88 @@
...
+        async function fetch() {
+            try {
+                setIsLoading(true)
+            } catch {
+                notification.error({ description: 'Failed to get shopping cart from backend' })
+            } finally {
+                setIsLoading(false)
+            }
+            fetch(`localhost:4000/user/${'userId'}/shoppingCart`).then(res => onChangeCart(res))
+        }
```
MehdiSaffar: We will be using axios instead of fetch as it is more developer-friendly. It will also allow us to provide a base url which is important when moving from local testing to production.

ecommerce/frontend/src/components/ShoppingCart/ShoppingCart.js
```diff
@@ -0,0 +1,88 @@
...
+        setItemCount(amountSum)
+    }
+
+    const onCheckout = () => {
+        if (itemCount == 0) {
+            alert('Empty cart')
+        } else {
+            alert('Total price: ' + totalPrice)
+        }
+    }
```
MehdiSaffar: it would be better to use Antd's notification or modal dialogs (they have an imperative interface similar to `alert`). 
Maybe disable the checkout button when shopping cart is empty? or have a whole different look to the page using Antd's `<Empty/>` component?

ecommerce/frontend/src/components/ShoppingCart/ShoppingCart.js
MehdiSaffar: Also when using inline formatting you should pass the variable directly, without using quotes like `${userId}`

MehdiSaffar: Oh and that fetch should actually be inside the try { }, no?
And did you try it? because I think fetch is calling itself recursively forever?

MehdiSaffar (changes requested):

MehdiSaffar:
I agree with @ozdenizdolu's comment about the look of the borders. It is better not use borders or at most a very faint color

ozdenizdolu:
I have merged my components into this one.

MehdiSaffar:
fixed key and margin-bottom errors

ozdenizdolu (approved):

PR #173, opened 2020-11-20
[FRONTEND] Profile Page

furkan-varol:
## Type of change

# Description

The profile page consists of 2 sections.
 1. Details Bar: Here, the user can change her password, see purchases she made, and go to the messages page. 
 2. Content Bar: This is the information about the account. The user can change and update the account information. This section is based on the original signup form we use to register new users.

# Screenshots
image

# Test results

No test

## Checklist:


ecommerce/frontend/src/components/UpdateProfileForm.js
```diff
@@ -0,0 +1,277 @@
...
+        userType: 'customer',
+        name: 'Furkan',
+        surname: 'Varol',
+        username: 'vaveyla',
+        email: 'furkan.varol@gmail.com',
+        phone: {
+            countryCode: '+4',
+            number: '8452391058',
+        },
+    }
```
MehdiSaffar: The initial values should a prop given to UpdateProfileForm

ecommerce/frontend/src/components/UpdateProfileForm.js
```diff
@@ -0,0 +1,277 @@
...
+            name="register"
+            onFinish={onFinish}
+            onFinishFailed={onFinishFailed}
+            initialValues={{
+                name: initialValues.name,
+                surname: initialValues.surname,
+                username: initialValues.username,
+                email: initialValues.email,
+                phone: initialValues.phone,
+            }}
```
MehdiSaffar: I believe you can pass `initialValues` directly to that prop?

ecommerce/frontend/src/components/UpdateProfileForm.js
```diff
@@ -0,0 +1,277 @@
...
+            }
+        }
+        return false
+    }
+
+    return (
+        <Form
+            {...formItemLayout}
+            form={form}
+            name="register"
```
MehdiSaffar: shouldn't name be `name="profile"` or something like that?

ecommerce/frontend/src/components/UpdateProfileForm.js
```diff
@@ -0,0 +1,277 @@
...
+                ) {
+                    return true
+                }
+                continue
+            } else if (initialValues[key] !== newValues[key]) {
+                return true
+            }
+        }
+        return false
+    }
```
MehdiSaffar: A better more general way would be to use one of Ramda's functionsl like `R.equals`  https://ramdajs.com/docs/#equals

We are already using Ramda in another feature btw.

ecommerce/frontend/src/components/pages/ProfilePage.js
```diff
@@ -0,0 +1,41 @@
...
+            </div>
+            <div className="profile-page-wrapper">
+                <div className="left-bar-profile-details">
+                    <ProfileDetails />
+                </div>
+                <div className="right-bar-profile-content">
+                    <ProfileContent />
+                </div>
+            </div>
+        </div>
```
MehdiSaffar: since `UpdateProfile` will get initial values as a prop you need to pass all the way down from this component too.
same thing for `ProfileDetails`

ecommerce/frontend/src/components/profile/ProfileDetails.js
```diff
@@ -0,0 +1,23 @@
...
+                    {' '}
+                    <Link to="/profile/change-password">Change Password</Link>{' '}
+                </li>
+                <li>
+                    {' '}
+                    <Link to="/profile/purchases">Purchase History</Link>{' '}
+                </li>
+                <li>
+                    {' '}
+                    <Link to="/profile/messages">Messages</Link>{' '}
```
MehdiSaffar: Please remove the `{' '}`

Also I would suggest using vertical AntD tabs for this instead. It will look much better in my opinion. https://ant.design/components/tabs/#components-tabs-demo-position

image

ecommerce/frontend/src/components/profile/ProfileDetails.less
```diff
@@ -0,0 +1,7 @@
+ul {
+    list-style-type: none;
+    display: flex;
+    flex-direction: column;
+    flex-wrap: wrap;
+    text-align: center;
+}
```
MehdiSaffar: This will change all unordered lists in the page which is bad. If you are going to use `ul`s you need to make a className for that list.

MehdiSaffar (changes requested):
I like the form but I don't like the black borders. 

Like I said in one of the comments, it would be better to use ant tabs for the left side.

Keep it up :)

ozdenizdolu (approved):
I think you have more or less done it. We can implement the "update info" later for another PR. Styling also can be finished later but you can do it if you want.

MehdiSaffar:
The style needs a lot of change but due to the time constraints we will have to merge this and leave that for another PR.

PR #212, opened 2020-12-04
[ANDROID] Product Endpoints & Switch Language Button

asliayk:
## Type of change

- [ ] Other

# Description

- I added three endpoints to ProductsAPI which are `products/homepage/{numberOfProducts}`, `product/{productId}` and `user/{userId}/listShoppingCart`.
- I added the necessary coroutine functions for these three endpoints, which are in CategoriesViewModel for now.
- According to the feedback, I added a switch language button to the top left corner of the login page. If the default language of the user is Turkish, EN button appears and when it's clicked, app is translated into English. If the default language of the user is English, TR button appears and when it's clicked, app is translated into Turkish. The button is reclickable, so user can return his/her default language.

# Screenshots

No screenshots, almost no change on design.

# Test results

No tests for now.

## Checklist:

- The coroutine functions can be moved to other classes, I wrote it there only for testing. You can test them by uncommenting the method calls in CategoriesFragment. We can move them after the SubcategoryFragment is done.
- I changed the base url with the Amazon url, we should use that url from now on.


ecommerce/android/Getflix/app/src/main/java/com/example/getflix/ui/viewmodels/CategoriesViewModel.kt
```diff
@@ -24,24 +24,62 @@ class CategoriesViewModel: ViewModel() {
...
+        }
+    }
+
+    fun getUserCartProducts(id: Int) {
+        val retrofit = Retrofit.Builder().baseUrl(MainActivity.StaticData.BASE_URL)
+                .addConverterFactory(GsonConverterFactory.create())
+                .build()
+                .create(ProductsAPI::class.java)
+        job = CoroutineScope(Dispatchers.IO).launch {
+            val response = retrofit.userCartProducts(id)
```
karalleyna: I think the code repeats itself where the retrofit object is created three times within different functions, you should move the creation of this project into the corresponding API service class. Also, it should be a singleton in order to prevent from creating completely same object again and again for each endpoint. It does not seem to be an efficient way to handle our requests

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/services/ProductsAPI.kt
```diff
@@ -1,16 +1,20 @@
...
 
 interface ProductsAPI {
 
-     @GET("products/homepage/3")
-     suspend fun getProducts(): Response<List<PModel>>
+     @GET("products/homepage/{numberOfProducts}")
+     suspend fun getProducts(@Path("numberOfProducts") numberOfProducts: Int): Response<List<PModel>>
 
+     @GET("product/{productId}")
+     suspend fun getProduct(@Path("productId") id: Int): Response<List<PModel>>
```
karalleyna: To be honest, I don't understand why you created a distinct class which uses the same api but different endpoints. Therefore, I think we should see eye to eye on which service,either GetflixApiService.kt or ProductsAPI.kt, we will be using and then unite all functions under this class instead of one to one match between an endpoint and a class. For example, if you add these functions into GetflixApiService.kt, then the issue with duplicate works in CategoriesViewModek.kt can also be handled, which means that we can kill two birds with one stone.

karalleyna (changes requested):

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/activities/MainActivity.kt
```diff
@@ -36,7 +29,7 @@ class MainActivity : AppCompatActivity() {
         var name = ""
         var sproducts = null
         val scategories = null
-        const val BASE_URL = "http://10.0.2.2:8000/"
+        const val BASE_URL = "http://ec2-18-189-28-20.us-east-2.compute.amazonaws.com:8000/"
```
karalleyna: I think you should move this variable into the class that proposes API requests since it is not related to the main activity.

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/ui/viewmodels/CategoriesViewModel.kt
asliayk: I think you haven't read my remarks section carefully. As I explained, I put them to this class only for testing because necessary classes are not ready(ex. subcategory). So, I didn't think whether writing 3 lines instead of 9 is more efficient or not, we can make necessary changes while moving. Our app has also other efficiency problems(your 500 lines of code in home fragment without custom view), so maybe we open a refactoring task next week and everyone makes necessary changes to her code.

karalleyna: Oh, please don't take it personally. :pensive: I think the efficiency does not come from the smaller number of codes but doing it at one shot if it is possible since, you know, the creation of the object takes time. I am not concerned with whether this could be done by shorter code or not. My purpose was how we,together, could implement functions,classes etc. as effective and robust as possible

asliayk: I'm not taking it personally, but I think we should make our reviews in a way that we respect our friends' effort. I don't understand why some efficiency problems "should be resolved" and others "would be better if resolved but not a must". You can check why a custom view is much more efficient, I think it's unnecessary to discuss it here.

karalleyna (approved):

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/services/ProductsAPI.kt
asliayk: moved to GetflixApiService now

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/activities/MainActivity.kt
asliayk: moved to GetflixAPIService now.

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/ui/viewmodels/CategoriesViewModel.kt
asliayk: used retrofit object in GetflixAPIService now.

PR #224, opened 2020-12-06
[ANDROID] Register View Model & Request for sign up endpoint

karalleyna:
## Type of change

# Description

1. Add the data models of SignUpCredentials and SignupResponse
2. Create signUp function for GetflixApiService.kt
3. Create the view model of the RegisterFragment.kt
4. Update the register fragment according to its view model, e.g adding live data observers to handle the events considering whether the user was able to sign up or not. 

# Screenshots

This pull request does not consist of visual changes.

# Test results

Tests are not provided since it is not necessary for the changes. 

## Checklist:


handesirikci:
as someone who has just learned the Kotlin language, altough right now i do not have a chance to test the code but i examined the commits and it seems correct

handesirikci:
as someone who has just learned the Kotlin language, altough right now i do not have a chance to test the code but i examined the commits and they do not seem problematic

asliayk (approved):
Nice to have a viewmodel for register fragment, I tested it and I think we can merge it now.

Next improvements for register page can be:
- Discussing which fields are mandatory for vendor registration and adding the necessary check functions
- Confirm password field check
- After register, maybe we can navigate back to login page or a dialog appears before navigating to home page. We should also discuss it.

Also please don't forget to pull before pushing to your branch.

PR #228, opened 2020-12-08
ANDROID]-Implement-Product-Page-Layout

mcanbayraktar:
## Type of change

# Description

I've designed product page based on project's color and design criteria.

# Screenshots

https://github.com/bounswe/bounswe2020group2/blob/master/images/Screen%20Shot%202020-12-08%20at%2014.03.26.png

# Test results

I could't test. 

## Checklist:

### DIFF
--- a/ecommerce/android/Getflix/.idea/render.experimental.xml
+++ b/ecommerce/android/Getflix/.idea/render.experimental.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="RenderSettings">
+    <option name="showDecorations" value="true" />
+  </component>
+</project>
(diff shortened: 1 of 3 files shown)


karalleyna:
Since it is a design, we should consider where our buttons should be placed in order to come in handy when one uses this mobile app, which colors should be used for a button to be clear about its desired action and functionality. I want you to think about these two scenarios. 
1. I am a customer who adds a product its favorites and I cannot be one hundred percent sure about whether I click the add favorites button while the add shopping cart is close to it, and vice versa.

2. Not every one has a perfect eyes to see everything as sharp as it is. So, I would not detect the black shopping cart item with dark purple background if I were an elder having an eye disease. (It should not be seen as an ageism. I want to provide just an example.)

asliayk:
We can't test it properly for now because the fragment is not ready yet, so let's merge it and @karalleyna makes necessary changes to it.

PR #230, opened 2020-12-08
[FRONTEND] User Reviews

### LINKED ISSUE #218: [FRONTEND] Implement user reviews component
Display user reviews in product page.

abdullahyldz:
## Type of change

- [ ] Other

# Description

User review component will be inside the product page. Component expects 'productId' and 'pageSize' as prop. It has a pagination at the bottom initialized at 1. Component makes a call to backend when current page is changed via a click event.

# Screenshots

Large Screen : image
Small Screen : image

# Test results

Manually tested.

## Checklist:

closes #218


MehdiSaffar:
I am reviewing this now.

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+                    },
+                    {
+                        comment: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor.',
+                        author: 'Hasan Kaya',
+                        date: 'December 11',
+                        rating: 3,
+                    },
+                ]
+                const urlStr = `/product/${productId}/review?pagesize=${pageSize}&currentpage=${currentPage}`
+                console.log(urlStr)
```
MehdiSaffar: for now completely fine since we don't know the endpoint's param, just leaving this here as a reminder for later to check the exact names of the params.

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+                        author: 'Jeff Einstein',
+                        date: 'December 23',
+                        rating: 4,
+                    },
+                    {
+                        comment: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor.',
+                        author: 'Hasan Kaya',
+                        date: 'December 11',
+                        rating: 3,
+                    },
```
MehdiSaffar: the `date` coming back from backend is likely going to be Date ISO8601 string not already formatted as `Month 11` or something.

it will be something like `2011-10-05T14:48:00.000Z`. parsing into a Date object (or using momentjs date library) will be required.

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+        console.log('Page changed to ', value)
+        setCurrentPage(value)
+    }
+
+    const Review = ({ review: { comment, author, date, rating } }) => {
+        return (
+            <div className="review-item-content">
+                <div className="rate-and-message">
+                    <div className="rate">
+                        <Rate defaultValue={rating} disabled={true}></Rate>
```
MehdiSaffar: Tip: shorter way is to simply write `disabled>`, by default it passes `true`

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+                            <p>{author}</p>
+                        </div>
+                    </Link>
+                    <div className="date">
+                        <p>{date}</p>
+                    </div>
+                </div>
+            </div>
+        )
+    }
```
MehdiSaffar: I would suggest moving `Review` component outside of the UserReview component

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+                    <Link to="/">
+                        <div className="author">
+                            <p>{author}</p>
+                        </div>
+                    </Link>
+                    <div className="date">
+                        <p>{date}</p>
+                    </div>
+                </div>
+            </div>
```
MehdiSaffar: styles applied to classNames are document-scoped not component-scoped. this means that then class names could conflict with other components. please prefix each class-name with `user-review-` for example

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+    const Review = ({ review: { comment, author, date, rating } }) => {
+        return (
+            <div className="review-item-content">
+                <div className="rate-and-message">
+                    <div className="rate">
+                        <Rate defaultValue={rating} disabled={true}></Rate>
+                    </div>
+                    <div className="message">
+                        <p>{comment}</p>
+                    </div>
```
MehdiSaffar: I believe there is no need to wrap comment additionally in a `<p>` tag, can you try without?

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+        fetch()
+    }, [currentPage])
+
+    const onPaginationChanged = value => {
+        // When page count is changed, current page is set
+        console.log('Page changed to ', value)
+        setCurrentPage(value)
+    }
+
+    const Review = ({ review: { comment, author, date, rating } }) => {
```
MehdiSaffar: To keep the naming consistent, name this UserReview (singular), and the top component UserReviews (plural)

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
...
+                    })}
+                </div>
+            </Spin>
+
+            <Pagination
+                className="review-results-pagination"
+                onChange={onPaginationChanged}
+                defaultCurrent={currentPage}
+                total={50}
+            />
```
MehdiSaffar: `total` should actually come from backend to show the current number of pages. Also need to pass `pageSize` to `Pagination` component. and instead of `defaultCurrent` you need to pass `current` since `defaultCurrent` only checks the values when it gets mounted.

ecommerce/frontend/src/components/UserReview.less
```diff
@@ -0,0 +1,51 @@
+.user-review {
+    display: flex;
+    flex-direction: column;
+    // gap: 10px;
+    padding: 2px 2%;
+    h1 {
+        font-size: xx-large;
+    }
+}
```
MehdiSaffar: I would suggest a white background and a `border: 2px gainsboro solid` to be consistent with the product header 
image

ecommerce/frontend/src/components/UserReview.js
```diff
@@ -0,0 +1,128 @@
+import React, { useEffect, useState } from 'react'
+import { Rate, Pagination, Spin } from 'antd'
+import { Link } from 'react-router-dom'
+import './UserReview.less'
+import { sleep } from '../utils'
+import { api } from '../api'
+
+export const UserReview = ({ productId = 1, pageSize = 10 }) => {
+    const [isLoading, setIsLoading] = useState(true)
+    const [currentPage, setCurrentPage] = useState(1)
```
MehdiSaffar: pay attention when passing the currentPage to backend since antd expects pages to be 1-indexed while backend expects 0-indexed, just letting you know.

MehdiSaffar (changes requested):
Overall solid work. Thank you for your effort :)

ecommerce/frontend/src/components/UserReview/UserReviews.js
```diff
@@ -1,11 +1,11 @@
...
-import { sleep } from '../utils'
-import { api } from '../api'
+import { Pagination, Spin } from 'antd'
+import './UserReviews.less'
+import { sleep } from '../../utils'
+import { api } from '../../api'
+import { UserReview } from './UserReview'
 
-export const UserReview = ({ productId = 1, pageSize = 10 }) => {
+export const UserReviews = ({ productId = 1, pageSize = 10, totalPage = 50 }) => {
```
MehdiSaffar: I would actually not have totalPage as a prop because the "total" is dependent on response from backend.

ecommerce/frontend/src/components/UserReview/UserReviews.js
```diff
@@ -20,44 +20,44 @@ export const UserReview = ({ productId = 1, pageSize = 10 }) => {
...
-                        date: 'December 11',
+                        date: '2011-10-05T14:48:00.000Z',
                         rating: 3,
                     },
                 ]
                 const urlStr = `/product/${productId}/review?pagesize=${pageSize}&currentpage=${currentPage}`
-                console.log(urlStr)
+                // console.log(urlStr)
                 // const { data: reviews } = await api.get(urlStr)
                 setReviews(tmp)
```
MehdiSaffar: The pagination info should come from backend so you could have a `setPagination({ current: ..., pageSize: ..., totalItems: ...})` for example

ecommerce/frontend/src/components/UserReview/UserReviews.js
```diff
@@ -120,8 +95,9 @@ export const UserReview = ({ productId = 1, pageSize = 10 }) => {
             <Pagination
                 className="review-results-pagination"
                 onChange={onPaginationChanged}
-                defaultCurrent={currentPage}
-                total={50}
+                current={currentPage}
+                defaultPageSize={pageSize}
+                total={totalPage}
```
MehdiSaffar, line 61: When you have the `pagination` state you can either spread the object directly as in

```js
<Pagination {...pagination} />
```
or just set each prop directly like
```js
<Pagination current={pagination.current} pageSize={pagination.pageSize} total={pagination.total} />
```

ecommerce/frontend/src/components/layout/Content.js
```diff
@@ -12,12 +12,14 @@ import { SignupPage } from '../pages/SignupPage'
...
 import { CategoryBar } from '../CategoryBar'
 import { CheckoutPage } from '../pages/CheckoutPage'
+import { UserReviews } from '../UserReview/UserReviews'
 
 export const Content = () => {
     return (
         <Layout.Content className="content">
             <CategoryBar />
             <Switch>
+                <Route path="/review" component={UserReviews} />
```
MehdiSaffar, line 22: did we decide to have the user reviews as a separate page or under the product header? I think you added this line just for easy testing right? if so then it is okay as long as you remember to remove this line later.

MehdiSaffar (changes requested):
Tried it with npm start looks good

Just an important change with regards to pagination

Good job :)

ecommerce/frontend/src/components/layout/Content.js
abdullahyldz: Reviews will reside under the Product page. That was only to test the component faster.

ecommerce/frontend/src/components/UserReview/UserReviews.js
```diff
@@ -1,66 +1,32 @@
...
-                        date: '2011-10-05T14:48:00.000Z',
-                        rating: 3,
-                    },
-                ]
-                const urlStr = `/product/${productId}/review?pagesize=${pageSize}&currentpage=${currentPage}`
-                // console.log(urlStr)
-                // const { data: reviews } = await api.get(urlStr)
-                setReviews(tmp)
-                await sleep(2000)
+                } = await api.get(`/review?product=${productId}&page_size=${pageSize}&page=${currentPage - 1}`)
```
MehdiSaffar: this is possible you could pass the params as an object I believe in the second parameter of api.get(...). both are fine as long as productId pageSize and currentPage will always have a non-null or non-undefined value.

ecommerce/frontend/src/components/UserReview/UserReview.js
```diff
@@ -15,7 +22,7 @@ export const UserReview = ({ review: { comment, author, date, rating } }) => {
             </div>
             <div className="review-author-and-date">
                 <Link to="/">
-                    <div className="review-author">{author}</div>
+                    <div className="review-author">{firstname.concat(' ', lastname)}</div>
                 </Link>
```
MehdiSaffar: cleaner way is `{firstname} {lastname}`

ecommerce/frontend/src/components/pages/ProductPage.js
```diff
@@ -6,6 +6,7 @@ import { api } from '../../api'
 import { sleep } from '../../utils'
 import { ProductHeader } from '../product/ProductHeader'
 import { notification, Spin } from 'antd'
+import { UserReviews } from '../UserReview/UserReviews'
 
 export const ProductPage = props => {
     const [product, setProduct] = useState()
```
MehdiSaffar, line 12: if you already have a `product` object why use `props.match.params` ?

MehdiSaffar (approved):
Good job :)

ecommerce/frontend/src/components/pages/ProductPage.js
abdullahyldz: Faster. `Product `is fetched using `productId` 
`UserReviews `needs only `productId` . There is no need to wait for `product `to be fetched.

ecommerce/frontend/src/components/UserReview/UserReviews.js
abdullahyldz: Do you mean sending the parameters as `body` ? 
I am fine as long as backend can handle request body parameters. I assumed they require path parameters. I will test that.

PR #242, opened 2020-12-12
[ANDROID] Filter/Sort Options & Refactoring

asliayk:
## Type of change

- [ ] Other

# Description

- I added the filter and sort options to `SubcategoryFragment`. To do that, I used a spinner for sorting and created `FilterOptionsFragment` from scratch. I used the same options with frontend for sort, and created the filter fragment with similar options with our project requirements for filter. I used MaterialAlertDialogBuilder to create a scrollable dialog on filter page. (The dialog opens by clicking the icons next to the titles)
- I added the `user/{userId}/shoppingCart` endpoint and its necessary coroutine function.
- I refactored the login endpoint. I got the user object in response and saved it to the main activity to be able to use it from everywhere on the app.
- I edited Hande's code to turn SubcategoryFragment into a grid layout (with 2 products in a row)
- I updated the categories with the list of categories which is shared, but I think it will be updated again with the coming endpoint.
- I deleted some of the classes which are unused and renamed some classes to avoid confusion.

# Screenshots

### FilterOptionsFragment
dialog

### Dialog on FilterOptionsFragment
dialog

### Grid Layout on SubcategoryFragment with Filter/Sort Options
grid

# Test results

No tests for now.

## Checklist:

- The options that the user chooses are saved on FilterOptionsFragment for now. I'm not sure how the endpoint for filter will be, so I didn't carry the filter data to the SubcategoryFragment. We can make the necessary changes after the endpoint implementation.
- Price range limits are 0-100 for now, we can discuss how to choose an upper limit


ecommerce/android/Getflix/app/src/main/java/com/example/getflix/models/ProductModel.kt
```diff
@@ -1,24 +1,22 @@
...
+    val image_url : String,
+    val total_rating : Int,
+    val rating_count : Int,
+    val stock_amount : Int,
+    val description : String,
+    val subcategory : String,
+    val category : String,
+    val brand : String,
+    val vendor : String
+) : Parcelable
```
karalleyna, line 22: I think we can use annotations in order to follow the naming convention of Kotlin. For example, `@SerializedName("stockAmount")`  instead of defining it as stock_amount. It's not a big issue, just want to mention it.

karalleyna (commented):
I reviewed all files which were updated and tested them. So, I believe these changes should be made to our project to align it with front end and will not create either a conflict or a bug. So, we can merge.

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/models/ProductModel.kt
asliayk: Yes, I learned that the json format of product endpoint has changed two days ago, but some endpoints still use the old version of it. So I think we can update it in the coming days, because I'm still not sure.

PR #255, opened 2020-12-17
B search 2

bcetin:


umutoksuz (approved):
Code seem fine, I also ran tests locally and passed.
I merged migrations to prevent errors in live db.

PR #260, opened 2020-12-19
[FRONTEND] F email verification

### LINKED ISSUE #251: [FRONTEND] Implement email verification (show necessary steps/notification in frontend)
Email verification needs to be implemented. 
I will at Frontend and by @bekir96 at Backend side. Details are given below.

abdullahyldz:
## Type of change

- [ ] Other

# Description

Two new features are implemented. 
First, a response page is created `EmailVerification.js` It shows response according to the validness of clicked URL.
Second, a verify button is placed under `ProfileDetails.js` in Profile page. 

# Screenshots

image
image
image

# Test results

EmailVerification.js is tested with @bekir96  from Backend.

## Checklist:

- [] I have commented my code, particularly in hard-to-understand areas

Verify button should be displayed if user is not verified. This information is not yet implemented by backend. Then, I will fix this. 

closes #251


ecommerce/backend/API/templates/acc_active_email.html
```diff
@@ -0,0 +1,6 @@
+{% autoescape off %}  
+Hi {{ user.username }},  
+Please click on the link to confirm your registration,
+http://127.0.0.1:5000/verify/{{ uid }}
+If you think, it's not you, then just ignore this email.  
+{% endautoescape %}
```
MehdiSaffar: I am not sure how to handle it but the ip address needs to be the production one when in production.. maybe have it depend on an environment variable somehow?

ecommerce/backend/API/utils/verify_email.py
```diff
@@ -0,0 +1,20 @@
...
+from API.utils.jwttoken import generate_access_token, generate_mail_token
+
+def email_send_verify(to_email, current_site, user):
+    mail_subject = 'Activate your account.'
+    message = render_to_string('acc_active_email.html', {
+                'user': user,
+                'domain': current_site.domain,
+                'url_path': 'email_verify',
+                'uid': generate_mail_token(user.pk),
+            })
```
MehdiSaffar: You probably need to add the domain ip as variable here for development/production environment.

ecommerce/frontend/src/components/EmailVerification.js
```diff
@@ -0,0 +1,69 @@
...
+                <p>Your activitation link is expired.</p>
+                <CloseCircleTwoTone twoToneColor="red" />
+            </div>
+        )
+    } else if (verificationState == 'Verified') {
+        return (
+            <div className="verify-message">
+                <p>Your account is already been verified. </p>
+                <CloseCircleTwoTone twoToneColor="red" />
+            </div>
```
MehdiSaffar: has* already been verified.

ecommerce/frontend/src/components/EmailVerification.js
```diff
@@ -0,0 +1,69 @@
...
+                console.error('failed to verify account', error)
+                setVerificationState('Invalid')
+            } finally {
+            }
+        }
+        fetch()
+    }, [])
+
+    return <div className="verify-message-box">{VerificationMessage(verificationState)}</div>
+}
```
MehdiSaffar: nitpick: would be better to use JSX notation for component as in: `<VerificationMessage ... />`

ecommerce/frontend/src/components/profile/ProfileDetails.js
```diff
@@ -1,10 +1,24 @@
...
+    const { user } = useAppContext()
+
+    const onVerify = () => {
+        async function verify() {
+            setIsLoading(true)
+            const { data } = await api.post('/user/verify', {})
+            setIsLoading(false)
+        }
+        verify()
+    }
```
MehdiSaffar: for a callback you don't need to create an inner async function you can simply write
```js
const onVerify = async () => {
   setIsLoading(true)
   ...
}
```

MehdiSaffar: also make sure you add try/catch/finally because if that api call fails for some reason then setIsLoading(false) would never happen leaving user with never ending loading spinner

MehdiSaffar: you would have the setIsLoading(false) in the finally scope. Also show an error notification

MehdiSaffar (changes requested):
Good job. I requested a couple changes. Also would it be possible to include the mock endpoints so that we can test how the frontend behaves to different answers?

Also maybe we can have a banner that appears below the category bar only if the user is not verified? 

image

using https://ant.design/components/alert/

Also for the verification message it would be nicer to use Antd components alreaady made for that https://ant.design/components/result/

image
image
image

abdullahyldz:
> Good job. I requested a couple changes. Also would it be possible to include the mock endpoints so that we can test how the frontend behaves to different answers?
> 
> Also maybe we can have a banner that appears below the category bar only if the user is not verified?

Yes, I have done it. It is kind of annoying but let's give it a try. 
Also, I added it in the Content.js, so it shows in the /verify endpoint, too. Showing 'verify account warning' to the user, when they are 'verifying their account' is funny

ecommerce/backend/API/templates/acc_active_email.html
abdullahyldz: @bekir96 it relates you.

ecommerce/backend/API/utils/verify_email.py
abdullahyldz: @bekir96 it relates you.

MehdiSaffar:
I am running the backend of this branch locally

I have two questions:

1. I see you are using `is_verified` field of user. Will /init be changed in this branch to add that field? or will that be in a seperate PR? (I feel like the first one is cleaner approach)
2. When I try to verify my account through the profile page button I get 405 Method Not Allowed error. Any idea why?
3. It would be better to have the banner be clickable because it is confusing for the error, there is no clear call to action.

abdullahyldz:
> 1. I see you are using `is_verified` field of user. Will /init be changed in this branch to add that field? or will that be in a seperate PR? (I feel like the first one is cleaner approach)
> 2. When I try to verify my account through the profile page button I get 405 Method Not Allowed error. Any idea why?
> 3. It would be better to have the banner be clickable because it is confusing for the error, there is no clear call to action.

1) I updated init and regular login methods.
2) Backend said they are working on some small changes, we test this after they push changes to live BE.
3) I am using Alert button the same way as the ant.desing page Somehow, it doesn't show up. 
```
<Alert
      message="Please verify your account."
      type="warning"
      action={
          <Spin spinning={isLoading}>
              <Space>
                  <Button size="small" type="ghost" onClick={onVerify}>
                      Done
                  </Button>
              </Space>
          </Spin>
      }
      closable
  />
```


I have created a new `EmailVerification.js` file to send alert. I had an issue with first entering the site that first current user is guest, then init call completes and user is set to my profile. 
Normally, I prevented the alert to pop up for guest users.
However, alert function did not show itself after the  user change. So, I used `currentUser `in the `EmailVerification.js`

abdullahyldz:
I and @bekir96 have completed the email verification system. We have made multiple manual tests together and corrected any misbehavior.

ecommerce/frontend/src/components/EmailVerification.js
```diff
@@ -0,0 +1,70 @@
...
+            </div>
+        )
+    } else {
+        return (
+            <div className="verify-message">
+                <Result icon={<LoadingOutlined />} title="Your account is being verified. Please wait..." />
+            </div>
+        )
+    }
+}
```
ozdenizdolu, line 41: I don't expect any changes here but I think a switch...case statement would be semantically more coherent with what is being done here.

ozdenizdolu (approved):
I read previous discussions between Mehdi & you. Looks like you did a great job overall. I tested only the frontend on my local (therefore it was a partial test.). It looks good to go. I will add an another comment If I fully test it on my local.

MehdiSaffar (approved):
Was merged to B-email-verification

PR #267, opened 2020-12-19
[ANDROID] Data Classes for New Endpoints & Vendor Fragments/Navigation

asliayk:
## Type of change

# Description

- For new endpoints, I created 7 data classes which are `AddressModel`, `BrandModel`, `CardModel`, `ReviewModel`, `VendorModel`, `ReviewedByModel` and `ExpirationDateModel`.
- I updated `ProductModel`, `CategoryModel`, and `SubcategoryModel` according to the improvements of endpoints.
- For vendor, I created 3 fragments which are `VendorHomeFragment`, `VendorProfileFragment`, and `MessagesFragment` , they are empty for now. I also added the necessary navigation by adding a different bottom navigation bar for vendor.
- I made some improvements to the register endpoint. I added necessary alerts to warn the user about character limits while entering necessary info, added confirm password check, and changed the navigation after a successful registration.
- I reordered some folders. I opened requests and responses folders under the service folder and moved request and response classes into these new folders.
- I added fragment animations to the navigations on the profile fragment for a better user experience.

# Screenshots

No screenshots needed.

# Test results

No tests for now.

## Checklist:

- There were some incompatibilities between some endpoints, that's why I didn't move the endpoint functions to the related fragments. So, the fragments with mock data are empty for now. When these endpoints are ready, we can remove all these mock datas.
- I tested my code with asliayk: vendor, and fatmay: customer accounts. I assumed id=20 indicates vendor for now. We can change it after the endpoint is updated.


karalleyna (approved):
I have tested and found that all additional data models do not have any problem in all our fragments and GetflixApiService. Also, the fragments are initialized for the scenario. In the following days, we can add their view models and complete their functions.

PR #276, opened 2020-12-20
[ANDROID] Creating CreditCard Fragment

### LINKED ISSUE #274: [ANDROID] Credit Card Info Page
User has to be able to see her saved credit cards and able to add new credit cards

handesirikci:
## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# Description

User can see her credit cards information by being navigated from profile page to Bank Accounts Info page, which shows all credit cards in a list and also enables user to add a new credit card by clicking on floating action button. After clicking, user enters the required information then adds the credit card to her profile.

# Screenshots

102009986-21465b80-3d4c-11eb-84bb-9c7524337bd3

## Checklist:


ecommerce/android/Getflix/app/src/main/java/com/example/getflix/models/CardModel.kt
```diff
@@ -2,12 +2,12 @@ package com.example.getflix.models
...
-        @SerializedName("owner_name") val owner_name : String,
-        @SerializedName("serial_number") val serial_number : Int,
-        @SerializedName("expiration_date") val expiration_date : ExpirationDateModel,
-        @SerializedName("cvc") val cvc : Int
+        @SerializedName("id") val id: Int,
+        @SerializedName("name") val name: String,
+        @SerializedName("owner_name") val owner_name: String,
+        @SerializedName("serial_number") val serial_number: Int,
+        @SerializedName("expiration_date") val expiration_date: String,
+        @SerializedName("cvc") val cvc: Int
```
karalleyna: I did not notice but our all data models serialized name are totally used wrong. I will change others as well. So, we can start this one. Could you change it this to the below code?
```suggestion
        val id: Int,
        val name: String,
        @SerializedName("owner_name") val ownerName: String,
        @SerializedName("serial_number") val serialNumber: Int,
        @SerializedName("expiration_date") val expirationDate: String,
        val cvc: Int
``` 
Because this annotation can be used to serialize a field with a different name instead of an actual field name in order to follow the naming convention

karalleyna (changes requested):
Overall, we can merge it after the minor change is done.

ecommerce/android/Getflix/app/src/main/java/com/example/getflix/models/CardModel.kt
asliayk: I didn't understand why it is wrong. Serialized names don't have to be different from the real variable names. I put the serialized names in case of any change in the names of the variables. So, it is not "totally wrong", is actually completely true and doesn't have to be updated.

asliayk (approved):
I tested the code, there was a bug while navigating from add credit fragment to back account info fragment, I fixed it by adding getItemCount function to the corresponding adapter. There was no other problem, I also resolved the conflicts and I think we can merge it now.

PR #385, opened 2021-01-16
[ANDROID] Empty Address & Credit Cart & Shopping Cart Bug

### LINKED ISSUE #379: [ANDROID] Bug due to empty shopping cart
### Preliminary Steps
When you are signing in as a guest, the view model of the shopping cart gives the following error after clicking the shopping cart icon at the bottom navigation bar.

### Screenshots
Ekran Resmi 2021-01-15 19 50 28

### Reasoning

It is not related to the type of the current user, the non-existence of products in the shopping cart list produces this error and then the app shuts down.  

### Priority

It needs to be fixed asap because it affects the functionality of the app as well as the process of new features to be added.

### Deadline
_**01.19.2021 @23.59**_

### LINKED ISSUE #384: [ANDROID] Bug happening after completing the order
### Preliminary Steps

1. Make sure that the address list of customer is empty
2. Go to the shopping cart page
3. Accept the current order

### Screenshots

https://user-images.githubusercontent.com/36455180/104805695-018ad300-57e3-11eb-8c60-707ec0f486b0.mov

### Reasoning

It is about functions calling endpoints assume the address list of the response is always empty.

### Solution

Functions regarding to endpoints should be fixed.

### Deadline
_**01.19.2021 @23.59**_

karalleyna:
## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# Description

You can find the explanation of these two bugs in #384 and #379. In order to fix them, I applied same procedure for both of them which is the followings :

1. Change both the type and return value of the necessary function in` GetflixApiService.kt`
2. Update data models which don't consider the odds of some empty fields in the response body.
3. Change the function calls and implementation in the view model in which bug occurs.
4. One of the common mistake that results in any of these bug is telling the compiler that this is always non empty anymore and same assumption for the user we cached in the `MainActivity.kt.` So, I removed them and wrote null checkers.

This proccess applied to the fragments that call functions, namely `getCustomerCards, getCustomerAddress, getCustomerCartPrice, and getCustomerAllProducts` of `GetflixApiService.kt`.

# Screenshots

https://user-images.githubusercontent.com/36455180/104807266-cc847d80-57ee-11eb-8552-9afa935a05bb.mov

## Checklist:


asliayk (approved):
I wasn't actually aware of the second bug, it's nice that they are both resolved now. I also tested it and I think it's ready to merge.

PR #401, opened 2021-01-19
[ANDROID] Register Form Update

### LINKED ISSUE #386: [ANDROID] Register Form Update
> Description

In sign up functionality, there are several missed fields that were not mentioned and have to be added to the form.

**_Deadline_: 20.01.2021, 23:59**

handesirikci:
## Type of change

# Description

In vendor sign up page, there are several missed fields that were not mentioned and have to be added to the form.
They were all about address information so i created an address part for the register page and classify the inputs that have to be taken from the user.

# Screenshots

https://user-images.githubusercontent.com/49978504/105080360-c1a54380-5aa1-11eb-9472-2d13fb6f908d.mp4

## Checklist:

There were more than one of input elements that correspond to the same information (phone, name, surname) in database so i had to add all the fields but it may not be useful for the user so maybe we can discuss it with backend team.


asliayk (changes requested):
I think there's a problem about changing the visibilities:
- When going from vendor to customer option, form sections of the address info dissappears(as expected), but Enter Address Info title is still visible. It should also be dissappeared.
- Another thing is when navigating to register page from login page, all the address form sections are visible for customer(which is default option). However, it shouldn't be visible.

It will be better if you fix these issues.

handesirikci:
Thanks for warning, you are right. I fixed those issues and i think it is OK now

PR #402, opened 2021-01-19
[BACKEND] B-create-product-list

### LINKED ISSUE #400: [BACKEND] Implement Create Product List endpoint
This endpoint generally work as post list name and respond product list id and status

bekir96:
## Type of change

# Description

This PR includes post lists endpoint. Request body should have list name :
image
And Response body is:
image
id represents pk of new created product list

# Test results

image

test for duplicate list name:
image

## Checklist:


ecommerce/backend/API/views/product_list.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@api_view(['POST'])
+@permission_classes([permissions.AllowAnonymous])
+def product_list_create(request):
+    jwt = authentication.JWTAuthentication()
+    user = jwt.authenticate(request=request)[0]
+
+    list_name = request.data.get("name")
+
```
omerfdeniz: If that user has a list with that name already, it will create a new one but with a different id. So you can return error or overwrite to it in that case.
 ```suggestion
    if ProductList.objects.filter(user_id=user.pk, name=list_name):
        return Response({ 'status': { 'successful': False, 'message': "User already has a list with that name."}})
```

omerfdeniz (dismissed):
Please prevent duplicate entries.

ecommerce/backend/API/views/product_list.py
bekir96: You are right, immediately it will be fixed

bekir96:
> Please prevent duplicate entries.

duplicate name error is fixed and suggestion is implemented

omerfdeniz (approved):
Now it is ready to merge, great work

umutoksuz (commented):
Clear code, well done

furkannane (approved):
Seems like the issues are solved, I see no problem as well.

PR #404, opened 2021-01-20
[BACKEND] B-get-product-list

### LINKED ISSUE #403: [BACKEND] Implement Get Product Lists endpoint
All product lists should be returned and product serializer is used for this purpose.

bekir96:
## Type of change

# Description

This PR includes get product lists endpoint. Response body is:

image
image

id represents pk of  product list

# Test results

image

## Checklist:


ecommerce/backend/API/views/product_list.py
```diff
@@ -0,0 +1,43 @@
...
+
+        product_list = ProductList(user_id=user.pk, name=list_name)
+        product_list.save()
+
+        return Response({'id': product_list.pk, 'status': { 'successful': True, 'message': "List is successfully created."}})
+
+    elif request.method == "GET":
+        response_product_list = []
+        product_lists = ProductList.objects.filter(user_id=user.pk)
+
```
omerfdeniz, line 34: Check the edge case if no such product_list exists.

```suggestion
if not product_lists:
    return Response({'status': {'successful': False, 'message':'No such list exists.'})

```

omerfdeniz (changes requested):
Left a comment to address before merge, other than that good work again

ecommerce/backend/API/views/product_list.py
umutoksuz: This is not a query function, this retrieves all lists of a user by user's id. So, if user has no lists created, it simply returns an empty list. Edge case may not be a problem.

umutoksuz (approved):
Looks good to me!

ecommerce/backend/API/views/product_list.py
furkannane: I think Umut is right. It is not a problem. Frontend may check if the list is empty and then they can print something like "you have no list".

furkannane (approved):
Nicely done, I saw no problem.

ecommerce/backend/API/views/product_list.py
bekir96: Actually, I traverse all product_lists with for loop, so if no item in it, response will return empty

bekir96 (commented):
the suggestion has been checked and there is no need for changes.

omerfdeniz (approved):

PR #406, opened 2021-01-20
[BACKEND] B-delete-product-list

### LINKED ISSUE #405: [BACKEND] Implement Delete Product List endpoint
id of product list should be given in url, and when delete list by this id,  all product should be deleted from list

bekir96:
## Type of change

# Description

This PR includes delete product lists endpoint. Response body is:
image

endpoint url is `/list/{id}`. Id field represents product list to be deleted and also represents pk of  product list.

# Test results

image

## Checklist:


ecommerce/backend/API/views/product_list.py
```diff
@@ -0,0 +1,59 @@
...
+
+@api_view(['DELETE'])
+@permission_classes([permissions.AllowAnonymous])
+def product_list_delete(request, list_id):
+    jwt = authentication.JWTAuthentication()
+    user = jwt.authenticate(request=request)[0]
+
+    product_list = ProductList.objects.filter(id=int(list_id))
+    if len(product_list) == 0:
+        return Response({'status': { 'successful': False, 'message': "This product list is invalid."}})
```
omerfdeniz: This might be more informative.
```suggestion
        return Response({'status': { 'successful': False, 'message': "User has no list with that id."}})
```

ecommerce/backend/API/views/product_list.py
```diff
@@ -0,0 +1,59 @@
...
+    else:
+        product_list.delete()
```
omerfdeniz, line 58: Since ProductListItem has CASCADE on delete of Product List, this is enough here.

omerfdeniz (changes requested):
Code is very good, just look at my comment on the message please, then I will approve

ecommerce/backend/API/views/product_list.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@api_view(['DELETE'])
+@permission_classes([permissions.AllowAnonymous])
+def product_list_delete(request, list_id):
+    jwt = authentication.JWTAuthentication()
+    user = jwt.authenticate(request=request)[0]
+
+    product_list = ProductList.objects.filter(id=int(list_id))
+    if len(product_list) == 0:
```
umutoksuz, line 55: I think you can delete product list even if it has no items in it. 
User may create a list, and then decides not to use this list and try to delete it.
I think we need to delete list in this case.

umutoksuz (changes requested):

ecommerce/backend/API/views/product_list.py
omerfdeniz: If statement there checks whether the requested product list actually exists not the number of items in it, so it is a valid check

furkannane: I think Umut is right. If statement checks if the product list has items to delete it but we might want to delete an empty list.

ecommerce/backend/API/views/product_list.py
furkannane: This message is not right. If I create a list with no items and want to delete it afterward, I will get this error message which doesn't seem right. Unless we are unable to create an empty product list, this is not an error case necessarily.

furkannane (changes requested):
Some of the little things should be fixed according to the feedback

ecommerce/backend/API/views/product_list.py
bekir96: @omerfdeniz @furkannane @umutoksuz  I will consider it according to your comments.

bekir96 (commented):
@omerfdeniz  says right, if-statement checks whether there is a list or not

ecommerce/backend/API/views/product_list.py
omerfdeniz: That if statement checks the # of product lists with that id, not the # of items in it, please look at from this perspective.

omerfdeniz (approved):
Great, ready to merge

furkannane (approved):
Code seems right now

umutoksuz (approved):

PR #423, opened 2021-01-22
[Backend] recommendation

furkannane:
## Type of change

# Description

I implemented the GET recommendation endpoint which returns some products to recommend to the customers according to their purchases. 

# Test results

1 test is added for the recommended products and it is passed.

## Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes


ecommerce/backend/API/views/recommendation.py
```diff
@@ -0,0 +1,81 @@
...
+        # get 5 products that are non-deleted 
+        products = Product.objects.filter(is_deleted=False)[:5]
+
+        for product in products:
+            serializer = ProductResponseSerializer(product)
+            serialized_product = serializer.data
+
+            recommended_products.append(serialized_product)
+
+    return Response({'products':recommended_products})
```
omerfdeniz, line 81: add status and message to the response please:

```suggestion
    return Response({'status': {'successful': True, 'message': "Successfully retrieved the recommendations."}, 'products':recommended_products})
```

ecommerce/backend/API/tests/test_recommendation.py
```diff
@@ -0,0 +1,76 @@
...
+        order = Order.objects.create(id=1, user_id=1)
+        purchase = Purchase.objects.create(id=1,amount=3,unit_price=22.11,status=0,purchase_date='2019-08-20 10:22:34+03',address_id=2,product=p,vendor=v,order_id=1)
+        order = Order.objects.create(id=2, user_id=1)
+        purchase = Purchase.objects.create(id=2,amount=3,unit_price=22.11,status=0,purchase_date='2019-08-20 10:22:34+03',address_id=2,product=p21,vendor=v,order_id=2)
+
+    @tag('recommend')
+    def test_1_recommendation(self):
+        response = self.client.get(reverse(recommend_products))
+
+        self.assertEqual(len(response.data["products"]), 2)
```
omerfdeniz, line 74: please also chech for 200 and status in the response:
```suggestion
        self.assertEqual(len(response.data["products"]), 2)
        self.assertEqual(response.status_code, 200)
        self.assertEqual(response.data["status"]["successful"], True)
```

ecommerce/backend/API/views/recommendation.py
omerfdeniz: It is better practice to use serializer with many field for a list serialization. Also please add the status and message to the response
```suggestion
        product_serializer = ProductResponseSerializer(products, many=True)
        return Response({'status': {'successful': True, 'message': "Successfully retrieved the recommendations."}, 'products': product_serializer.data})
        
```

omerfdeniz (changes requested):
Address the response type and serialization please

ecommerce/backend/getflix/urls.py
```diff
@@ -49,5 +49,6 @@
     path('customer/orders', views.order.customer_order),
     path('messages', views.message.manage_messages),
     path('notifications', views.notification.manage_notifications, name="notifications"),
-    path('lists', views.product_list.product_list_create)
+    path('lists', views.product_list.product_list_create),
+    path('recommendation', views.recommendation.recommend_products)
 ]
```
bekir96: can you resolve conflict in urls.py

bekir96 (changes requested):
The master may have changed, you will have to look at conflicts.

ecommerce/backend/API/views/recommendation.py
```diff
@@ -0,0 +1,81 @@
...
+from ..utils.order_status import OrderStatus
+
+@api_view(['GET'])
+@permission_classes([permissions.IsCustomerUser])
+def recommend_products(request):
+
+    # get user
+    user = request.user
+    user_id = request.user.id
+
```
umutoksuz, line 22: Here if no token is provided in request header, and user will be none. You should check for that and return an error in case

umutoksuz (changes requested):
Some changes requrested

ecommerce/backend/API/views/recommendation.py
furkannane: Added status and message.

ecommerce/backend/API/tests/test_recommendation.py
furkannane: Done

PR #434, opened 2021-01-23
[FRONTEND] Recommendations

### LINKED ISSUE #376: [FRONTEND] User specific Recommended Products
A component to hold the recommended products needs to be implemented.
Decisions to be made:
1) Design (An option: HorizontalProductList that used on HomePage for Best Sellers)
2) Where to display the component.
3) Depending on Item 2; the format we recommend products.

furkan-varol:
## Type of change

- [ ] Other

# Description
A horizontal list that displays a list of products personalized for each user depending on their previous orders for the Home Page. Here, the previously implemented HorizontalProductList component is used.

# Screenshots

image

# Test results
Backend integration is done and tested.

## Checklist:

Backend endpoint `/recommendation` returns the response after a noticeable delay. I do not know the exact reason, will try to fix this.

Closes #376.


ecommerce/frontend/src/components/Recommendations.js
```diff
@@ -0,0 +1,47 @@
...
+            notification.error({ description: 'Failed to load recommendations' })
+        } finally {
+            setIsLoading(false)
+        }
+    }
+
+    return (
+        <Spin spinning={isLoading}>
+            <div className="recommendations-wrapper">
+                <h1 className="recommendations-header">Recommended Products For You</h1>
```
ozdenizdolu: Maybe you can write something more personalized like "We have selected these products for you Özdeniz!". I don't know the best way to write this but I think you might try to include the first name of the user. Feel free to not do it.

ozdenizdolu (approved):
Other than my comment on the personalized title, it looks good and seems to be working fine. Good job.

ecommerce/frontend/src/components/Recommendations.js
```diff
@@ -0,0 +1,47 @@
...
+                <div className="recommendations-content">
+                    {products.map(product => (
+                        <ProductCard key={product.id} product={product} />
+                    ))}
+                </div>
```
MehdiSaffar, line 44: In case the user is new and they don't have any recommendations, it would be nice to either hide this entire component or show something like "We currently do not have recommendations for your". I would prefer the first option.

MehdiSaffar (approved):
Looks good to me.

ecommerce/frontend/src/components/Recommendations.js
furkan-varol: I discussed this with @furkannane, the response always has >0 # of products. Even when the customer has no previous orders.

ecommerce/frontend/src/components/Recommendations.js
furkan-varol: I have added the name of the user to the header. 
image

PR #461, opened 2021-01-24
[B] Vendor Profile Implemented

### LINKED ISSUE #416: [B] Implement Vendor Update profile
Following this template:
```
2. vendor/profile (PUT) [Token required] (Delete image)
   
   - Request
   
      ```json 
      {
          "title": "X shop",
          "description": "description",
          "image": "base 64 encoded image"
      }
      ```
   ```
   
   - Response
   
      ```json 
      {
          "status": {
              "successful": true,
              "message": "Profile updated successfully"
          }
      }
   ```

```

bcetin:
## Type of change

# Description

I have implemented the vendor/profile endpoint which changes vendor title, description and image in the database.

## Checklist:

# Test Results
```
Using existing test database for alias 'default'...
System check identified no issues (0 silenced).
.
----------------------------------------------------------------------
Ran 1 test in 5.065s

OK
```


ecommerce/backend/API/views/vendor.py
```diff
@@ -14,3 +14,20 @@ def vendor_details(request,vendor_id):
...
+    image.save()
+    image_url="/image/"+str(image.pk)
+    Vendor.objects.filter(user_id=vendor.id).update(title=request.data["title"],description=request.data["description"],image_url=image_url)
+    return Response({
+          "status": {
+              "successful": "true",
+              "message": "Profile updated successfully"
+          }
+      }
+)
```
bekir96: Can you add test for this endpoint?

ecommerce/backend/API/views/vendor.py
```diff
@@ -14,3 +14,20 @@ def vendor_details(request,vendor_id):
     if vendor is None:
         return Response(status=status.HTTP_400_BAD_REQUEST)
     return Response({'title':vendor.title,'image_url':vendor.image_url,'first_name':vendor.first_name,'last_name':vendor.last_name,'last_name':vendor.last_name,'total_rating':vendor.total_rating,'rating_count':vendor.rating_count,'description':vendor.description})
+
+@api_view(['PUT'])
+@permission_classes([permissions.IsVendorUser]) 
+def vendor_profile(request):
+    vendor = User.objects.filter(username=request.user).first()
```
umutoksuz: I am suspicious about this line of code. request.user is a user object. Checking its equality with username would fail I think. 

vendor = Vendor.objects.filter(user_id=request.user.id).first()

would be appropriate

bekir96 (changes requested):
You should add test to check endpoint.

ecommerce/backend/API/views/vendor.py
```diff
@@ -14,3 +14,20 @@ def vendor_details(request,vendor_id):
...
+    img_array = base64.b64decode(request.data["image"])
```
umutoksuz: You should check if no image file is sent from frontend. 
For example, vendor already have a profile picture and wants to only update his/her description. In this case, this code forces vendor to send same picture again to backend.

If image is null, we should not change image_url field.

umutoksuz (changes requested):
Can you implement at least one unit test.

ecommerce/backend/API/views/vendor.py
bcetin: There is a test added in test_vendor.

ecommerce/backend/API/views/vendor.py
```diff
@@ -14,3 +14,23 @@ def vendor_details(request,vendor_id):
...
+def vendor_profile(request):
+    vendor = User.objects.filter(id=request.user.id).first()
+    if "image" in request.data:
+        Vendor.objects.filter(user_id=vendor.id).update(title=request.data["title"],description=request.data["description"])
+    else:
+        img_array = base64.b64decode(request.data["image"])
+        image = Image(image=img_array)
+        image.save()
+        image_url="/image/"+str(image.pk)
+        Vendor.objects.filter(user_id=vendor.id).update(title=request.data["title"],description=request.data["description"],image_url=image_url)
```
bekir96: You control if image in request.data, but in else statement, you take request.data["image"].  There is a conflict.

bekir96 (changes requested):
Please consider conflict situation mentioned in review.

ecommerce/backend/API/views/vendor.py
bcetin: Fixed

bekir96 (approved):
All sufficient changes are implemented, you can merge

